PR #1, opened 2022-12-27
Change KDF domain separation to be > 32 chars

wussler:
NIST.SP.800-185-2 specifies:
> Applications of this Recommendation shall not select an input key, K, whose length is less than their required security strength. Guidance for cryptographic algorithm and key-size selection is available in [4].

I would raise the length of our domain separation string to > 32 chars for conformance with the specification. Also many implementations implement this strictly, so this ensures easy reuse of existing implementations.

Furthermore I'd drop the `V5` from the string. I don't think we want to customize the string based on the PKESK version (we also don't for ECDH, that can be used across v4 and v5)

### DIFF
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -808,7 +808,7 @@ construction is a one-step key derivation function compliant to
     //   multiKeyCombine(eccKeyShare, kyberKeyShare, fixedInfo)
     //   Input:
     //   domSeparation - the UTF-8 encoding of the string
-                         "OpenPGPV5CompositeKDF"
+                         "OpenPGPKyberCompositeKeyDerivation"
     //   counter - a 4 byte counter set to the value 1
     //   eccKeyShare - the ECC key share encoded as an octet string
     //   kyberKeyShare - the Kyber key share encoded as an octet string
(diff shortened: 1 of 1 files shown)


falko-strenzke:
I'm not sure I understand. It is the KMAC *key* that needs to have a sufficient length. The domain seperation just needs to be unique.

wussler:
As far as I understand we're effectively using `domSeparation` as a KMAC key

falko-strenzke:
OK, I see, it really seems to be an sublte inconsistency between the NIST specification of the one step key deriviation based on KMAC and this restriction of KMAC regarding the key length. So I think your suggestion is the correct solution.

falko-strenzke (approved):

fluppe2:
We can extend the length of the domain seperation string, but I wouldn't put "Kyber" in there for two reasons:

a) The name "Kyber" will eventually turn into "MLWE". I suggest to keep the amount of necessary code changes when we align to the forthcoming NIST standards as small as possible. This is somehow an analoguous argument as to not include version information.

b) The composite KDF does not care if it is feeded with ECC, Kyber or somoething else. Therefore I suggest to keep the domain seperation generic. e.g. if the CFRG decides to standardize NTRU in the future, the KDF will receive maybe an ECC || NTRU value.

I would suggest, instead of putting "OpenPGPKyberCompositeKeyDerivation" we should do something like "OpenPGPCompositeKeyDerivationFunction".

wussler:
So @fluppe2 you would keep the same hash for Kyber or other algorithms?

I am happy with the proposal `OpenPGPCompositeKeyDerivationFunction`

fluppe2:
@wussler : Yes, I would prefer the domain separation to specify a generic composer, so that potential future PQC-extensions for OpenPGP could adopt this. The fixedInfo field contains the specific algorithm information anyway.

falko-strenzke:
> @wussler : Yes, I would prefer the domain separation to specify a generic composer, so that potential future PQC-extensions for OpenPGP could adopt this. The fixedInfo field contains the specific algorithm information anyway.

Makes sense, is fine for me, too

PR #36, opened 2023-03-16
IND-CCA2 for ECC-KEM and KEM Combiner

fluppe2:
What has been added or changed:

a) Pointed to the Bertoni/Daemen/Peters/Assche paper "On the Indifferentiability of the Sponge Construction" for Keccak since this has a stable DOI
b) Defined IND-CCA2 secure KEM combiner according to the Giacon/Heuer/Poettering paper "KEM Combiners" and added security considerations for this
c) Aligned x25519kem and x448kem to RFC7748, splitted the description to be more readable
d) Defined IND-CCA2 secure ECC-KEMs according to the Cramer/Shoup paper "Design and Analysis of Practical Public-Key Encryption Schemes Secure against Adaptive Chosen Ciphertext Attack" and added security considerations for this
e) Removed the words "native", "masking", "clamping" concerning x25519, x448 as they are nowhere defined and might not be comprehensible, instead defined to simply follow the encodings of RFC7748
f) Added "oBits=256" into the call of "multiKeyCombine" of the composite KEM encryption and decryption procedure as the KEK needs to be specified

@wussler: Are you okay with c) and e)?

@ahuelsing: Are you okay with a)? Can you check b)? Can you check also d)?

@falko-strenzke, @TJ-91: Are you okay with f)? What did you guys implement? Did you implicitely assume this?

### DIFF
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -190,24 +190,6 @@ informative:
     seriesinfo:
       NIST Special Publication 800-186
 
-  SPONGE:
-    target: https://keccak.team/files/CSF-0.1.pdf
-    title: Cryptographic sponge functions
-    author:
-      -
-        ins: G. Bertoni
-        name: Guido Bertoni
-      -
-        ins: J. Daemen
-        name: Joan Daemen
-      -
-        ins: M. Peters
-        name: Michael Peters
-      -
-        ins: G. Assche
-        name: Gilles van Assche
-    date: January 2011
-
   SEC1:
     target: https://secg.org/sec1-v2.pdf
     title: "Standards for Efficient Cryptography 1 (SEC 1)"
@@ -301,12 +283,58 @@ informative:
       date: 2021-06-10
 
   draft-driscoll-pqt-hybrid-terminology:
-    target: https://datatracker.ietf.org/doc/html/draft-driscoll-pqt-hybrid-terminology-01
+    target: https://datatracker.ietf.org/doc/html/draft-driscoll-pqt-hybrid-terminology
     title: Terminology for Post-Quantum Traditional Hybrid Schemes
     author:
       -
         ins: F. Driscoll
-    date: October 2022
+        name: Florence Driscoll
+    date: March 2023
+
+  GHP18:
+    target: https://doi.org/10.1007/978-3-319-76578-5_7
+    title: KEM Combiners
+    date: 2018
+    author:
+      -
+        ins: F. Giacon
+        name: Federico Giacon
+      -
+        ins: F. Heuer
+        name: Felix Heuer
+      -
+        ins: B. Poettering
+        name: Bertram Poettering
+
+  BDPA08:
+    target: https://doi.org/10.1007/978-3-540-78967-3_11
+    title: On the Indifferentiability of the Sponge Construction
+    author:
+      -
+        ins: G. Bertoni
+        name: Guido Bertoni
+      -
+        ins: J. Daemen
+        name: Joan Daemen
+      -
+        ins: M. Peters
+        name: Michael Peters
+      -
+        ins: G. Assche
+        name: Gilles van Assche
+    date: 2008
+
+  CS03:
+    target: https://doi.org/10.1137/S0097539702403773
+    title: Design and Analysis of Practical Public-Key Encryption Schemes Secure against Adaptive Chosen Ciphertext Attack
+    author:
+      -
+        ins: R. Cramer
+        name: Ronald Cramer
+      -
+        ins: V. Shoup
+        name: Victor Shoup
+    date: 2003
 
 --- abstract
 
@@ -428,8 +456,8 @@ schemes.
 
 Curve25519 and Curve448 are defined in [RFC7748] for use in a Diffie-Hellman
 key agreement scheme and defined in [RFC8032] for use in a digital signature
-scheme. For Curve25519 this specification adapts the encoding of objects in
-native format in contrast to [I-D.ietf-openpgp-crypto-refresh].
+scheme. For Curve25519 this specification adapts the encoding of objects as
+defined in [RFC7748] in contrast to [I-D.ietf-openpgp-crypto-refresh].
 
 ### Generic Prime Curves
 
@@ -517,9 +545,9 @@ Even though the zero point, also called the point at infinity, may occur as a
 result of arithmetic operations on points of an elliptic curve, it MUST NOT
 appear in any ECC data structure defined in this document.
 
-Furthermore, when performing the explicitly listed multiplications in
-{{x25519-x448-kem}} or {{ecdh-kem}} it is REQUIRED to perform all checks,
-clamping, or masking mandated from the relative elliptic curve specification.
+Furthermore, when performing the explicitly listed operations in
+{{x25519-kem}}, {{x448-kem}} or {{ecdh-kem}} it is REQUIRED to follow the
+security advisory mandated from the relative elliptic curve specification.
 
 
 # Supported Public Key Algorithms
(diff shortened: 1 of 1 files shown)


draft-wussler-openpgp-pqc.md
```diff
@@ -517,9 +545,9 @@ Even though the zero point, also called the point at infinity, may occur as a
 result of arithmetic operations on points of an elliptic curve, it MUST NOT
 appear in any ECC data structure defined in this document.
 
-Furthermore, when performing the explicitly listed multiplications in
-{{x25519-x448-kem}} or {{ecdh-kem}} it is REQUIRED to perform all checks,
-clamping, or masking mandated from the relative elliptic curve specification.
+Furthermore, when performing the explicitly listed operations in
+{{x25519-kem}}, {{x448-kem}} or {{ecdh-kem}} it is REQUIRED to follow the
+security advisory mandated from the relative elliptic curve specification.
```
falko-strenzke, line 550: I don't think "security advisory" is entirely covering what is intended to be conveyed by the previous formulation. It would rather say:
```suggestion
specification and security advisory mandated from the relative elliptic curve specification.
```

falko-strenzke (approved):
As far as I can evaluate the MR, it is fine. But Andreas should still approve the formal claims Stavros suggested.

draft-wussler-openpgp-pqc.md
```diff
@@ -688,7 +721,9 @@ lengths.
 | ECDH public key        | 65 octets of SEC1-encoded public point                 | 97 octets of SEC1-encoded public point                 |
 | ECDH secret key        | 32 octets big-endian encoded secret scalar             | 48 octets big-endian encoded secret scalar             |
 | ECDH ephemeral         | 65 octets of SEC1-encoded ephemeral point              | 97 octets of SEC1-encoded ephemeral point              |
-| Key share              | 32 octets of `X` coordinate decoded from SEC1 encoding | 48 octets of `X` coordinate decoded from SEC1 encoding |
+| ECDH share             | 65 octets of SEC1-encoded shared point                 | 97 octets of SEC1-encoded shared point                 |
+| Key share              | 32 octets                                              | 48 octets                                              |
+| Hash                   | SHA3-256                                               | SHA3-384                                               |
```
wussler, line 726:
```suggestion
| Key share              | 32 octets                                              | 64 octets                                              |
| Hash                   | SHA3-256                                               | SHA3-512                                               |
```

draft-wussler-openpgp-pqc.md
```diff
@@ -706,29 +741,57 @@ The placeholder `eccKem` has to be replaced with the specific ECC-KEM from the
...
+
+ 1. Generate an ephemeral key pair {`v`, `V`} via `V = X448(v,5)`
+
+ 2. Compute the shared coordinate `X = X448(v, R)` where `R` is the public key
+    `eccPublicKey`
+
+ 3. Set the output `eccCipherText` to `V`
+
+ 4. Set the output `eccKeyShare` to the first 56 octets of `SHA3-512(X ||
+    eccCipherText)`
```
wussler, line 786: I'm personally not the biggest fan of this change. Given that the length of the ciphertexts is guaranteed by protocol construction I would not add a hashing here, that's expensive!

Also, I find silly truncating the hash, when anyway it'll end up in a KDF.

draft-wussler-openpgp-pqc.md
```diff
@@ -1405,31 +1474,51 @@ computer attacks.
...
+More precisely, for a given capacity `c` the indifferentiability proof shows
+that assuming there are no weaknesses found in the Keccak permutation, an
+attacker has to make an expected number of `2^(c/2)` calls to the permutation
+to tell KMAC from a random oracle. For a random oracle, a difference in only a
+single bit gives an unrelated, uniformly random output. Hence, to be able to
+distinguish a key `K`, derived from shared keys `K1` and `K2` (and ciphertexts
+`C1` and `C2`) as
 
-    K = KMAC(domainSeparation, counter || K1 || K2 || fixedInfo,
+    K = KMAC(domainSeparation, counter || K1 || C1 || K2 || C2 || fixedInfo,
```
wussler, line 1505: Isn't this inconsistent with the hashing done above?
I would really prefer this version, as we can simply argue it's secure and more performant

wussler (commented):
Hi @fluppe2, thanks for this very good PR!

I like the change in (c) and (e), but am not happy with the pre-hashing of the secrets with the ciphertext. I would rather include the ciphertexts directly in the KDF as mentioned in the security considerations.

Also, FYI SHA3-384 was explicitly excluded from the crypto-refresh, so I'd abstain from using it here

draft-wussler-openpgp-pqc.md
ahuelsing: I agree that there is potentially an optimization possible. However, we have seen that the original draft was actually insecure (the ECC part was not CCA2 secure). I would adopt the change. This definitely gives you security in a modular way.  In my opinion we can optimize it if we manage to directly prove that the optimization is also secure.

draft-wussler-openpgp-pqc.md
```diff
@@ -1001,18 +1078,16 @@ The algorithm-specific public key is this series of values:
  - A fixed-length octet string representing an EC point public key, in the
    point format associated with the curve specified in {{ecc-kem}}.
 
- - A fixed-length octet string containing the Kyber public key in native
-   format, whose length depends on the algorithm ID as specified in
-   {{tab-kyber-artifacts}}.
+ - A fixed-length octet string containing the Kyber public key in, whose length
```
ahuelsing, line 1081:
```suggestion
 - A fixed-length octet string containing the Kyber public key, whose length
```

draft-wussler-openpgp-pqc.md
```diff
@@ -1405,31 +1474,51 @@ computer attacks.
...
+The proposed construction in {{kem-key-combiner}} preserves IND-CCA2 of any of
+its ingredient KEMs, i.e. the newly formed combined KEM is IND-CCA2 secure as
+long as at least one of the ingredient KEMs is. Indeed, the above stated
+indifferentiability from a random oracle qualifies Keccak as a split-key
+pseudorandom function as defined in {{GHP18}}. That is, Keccak behaves like a
+random function if at least one input shared secret is picked uniformly at
+random. Our construction can thus be seen as an instantiation of the IND-CCA2
+preserving Example 3 in Figure 1 of {{GHP18}}, up to some reordering of input
+shared secrets and ciphertexts. In the random oracle setting, the reordering
+does not disturb the arguments in {{GHP18}}.
```
ahuelsing, line 1520:
```suggestion
does not influence the arguments in {{GHP18}}.
```

ahuelsing (approved):
Looks good. I only suggested a few editorial changes and answered comments

draft-wussler-openpgp-pqc.md
wussler: Sure, but then it should be expressed as 
```
K = KMAC(domainSeparation, counter || H( K1 || C1 ) || H( K2 || C2 ) || fixedInfo,
```
In this consideration we don't account for the additional hashing step

draft-wussler-openpgp-pqc.md
wussler: @ahuelsing don't get me wrong, I am happy with `X ||  eccCipherText`, but I would remove the `SHA-3-512` here (and in the other algorithms too)

draft-wussler-openpgp-pqc.md
ahuelsing: This is possible, but not necessary (from a security point of view).

draft-wussler-openpgp-pqc.md
ahuelsing: And then you do not have a secure KEM.

draft-wussler-openpgp-pqc.md
ahuelsing: If you use that version of the combiner, but indeed, then you have an unnecessary hashing for Kyber. Moreover, do you have a KEM description which defines an insecure KEM (which should at least be highlighted in that section in this case). The danger that I see is that extensions to new algorithms may render the whole thing insecure if we are not explicit where we do require what... (and in the previous version of the draft this was not clear at all)

wussler: So basically what this construction wants is the following?
```K = KMAC(domainSeparation, counter || H( SS1 || C1 ) || C1 || H( SS2 || C2 ) || C2 || fixedInfo,```
Since K1 is already defined here as H( SS1 || C1 )?

Even if it's not necessary from the security standpoint, I'd like to keep this sentence as close as possible to what we're doing in order to cause less confusion

File: draft-wussler-openpgp-pqc.md [hunk 7]
```diff
@@ -1069,27 +1144,26 @@ Algorithm ID reference | Curve           | Field size | Public key | Secret key
...
 given in Figure 4 in [DILITHIUM-Subm], where `sk` is the Dilithium private key
-in native format and `M` is the data to be signed. OpenPGP does not use the
-optional randomized signing given as a variant in the definition of this
-function, i.e. `rho' := H(K || mu)` is used. The signing function returns the
-Dilithium signature in native format. That is, to sign with Dilithium the
-following operation is defined:
+and `M` is the data to be signed. OpenPGP does not use the optional randomized
+signing given as a variant in the definition of this function, i.e. `rho' :=
+H(K || mu)` is used. The signing function returns the Dilithium signature. That
+is, to sign with Dilithium the following operation is defined:
```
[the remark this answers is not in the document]
fluppe2: I denoted the basepoints of Curve25519 and Curve448 in their abstract form following the notation of RFC7748 as U(P). See #[f2af0c5](https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/36/commits/f2af0c5023d5b0cfd30207491c859ca205757e93)

fluppe2:
Let me try to explain my approach with an IND-CCA2 version ofthe ECC-KEMs in conjunction with the KEM combiner.
- I tried to build this in a modular way, i.e. have CCA2-secure ingredient KEMs and then feed the ciphertexts into the KEM combiner so that the KEM combiner preserves their CCA2-security.
- The hashing in the ECC-KEMs ensures CCA2-security, the hashing is necessary if you want to achieve CCA2-security for the ECC-KEM itself. Only concatenation is not enough for this purpose as an attacker would control changes to the concatenation bitwise, the hashing ensures that changes propagate.
- Having the hashing step in the ECC-KEMs, I think it is enough to simply input the K1 || C1 || K2 || C2 into the combiner as the lengths of all key shares and ciphertexts are constant depending only on the specific parameters of the ingredient KEMs. We do not have to introduce an additional H(K1||C1) || H(K2||C2) in the combiner.
- If we decide to remove the hashing step in the ECC-KEMs we would have to adopt our security considerations,trying to argue that the combined construction as a whole is (maybe) CCA2-secure, then I would rather suggest to have the KEM combiner processing H(K1||C1) || H(K2||C2), since if K1, C1 come from an ECC-KEM, the hashing step for ECC-KEM would then be performed inside the combiner. Maybe that would allow a security argument that while the ECC-KEM building blocks are only IND-CPA the construction as a whole would (maybe) ensure IND-CCA2. I don't know. If K2,C2 come from Kyber then the additional hashing might be superfluous but relieves an implementer from length checks so might be beneficial.

draft-wussler-openpgp-pqc.md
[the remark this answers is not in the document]
wussler: It's defined in figure 4, referenced few lines above. I would be OK with not adding a second reference.

wussler:
@fluppe2 I addressed the comments we had today in the call, please review the latest changes and when you're OK we can merge :rocket:

fluppe2:
I would say the mentioning of rho is fine for now, I think it is still a moving target how we reference the PQ specs, since we are waiting for the final standards.

PR #42, opened 2023-03-27
Align v3 PKESK to Crypto Refresh

### LINKED ISSUE #35: Restrictions regarding combinations of legacy and "PQC" recipients
Johannes pointed out that we have the following restriction:

- Since our draft restricts PQC algos to be used only with v5 PKESK (which we do not really need for the implicit rejection with AEAD through the associated v2 SEIPD because the AES Key Wrap has integrity protection [RFC3394](https://www.rfc-editor.org/rfc/rfc3394#section-2.2.3) which already provides implicit rejection ), a sender cannot send a message to both a legacy recipient that doesn't support v5 PKESK and a recipient for which the sender only knows PQC keys.
- The cryptorefresh allows in this case downgrade to v3 PKESK / v1 SEIPD.
- We have the following alternatives:
  - We can stick to this restriction,
  - or we can drop the requirement which then would allow the same downgrade as in the crypto-refresh.
    - for this we would have to add padding bytes in case of the v3 PKESK (because here the algo ID is also included). This is the same issue as discussed [here](https://gitlab.com/openpgp-wg/rfc4880bis/-/issues/157) for the crypto-refresh and the solution defined there should be used in our case, too.

This needs a decision.

TJ-91:
This is a suggestion to align the new PQC encryption algorithms with X25519/X448 in the Crypto Refresh in regards to the PKESK packets

I also fixed  occurances of "symmetric key" and replaced it by "wrapped session key"

Closes #35

### DIFF
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -1066,13 +1066,13 @@ scheme is as follows:
 
 The algorithm-specific fields consists of:
 
- - The one-octet algorithm identifier, if it was passed (in the case of a v3 PKESK packet).
-
  - A fixed-length octet string representing an ECC ephemeral public key in the
    format associated with the curve as specified in {{ecc-kem}}.
 
  - A fixed-length octet string of the Kyber ciphertext, whose length depends on
    the algorithm ID as specified in {{tab-kyber-artifacts}}.
+
+ - The one-octet algorithm identifier, if it was passed (in the case of a v3 PKESK packet).
 
  - A variable-length field containing the wrapped session key:
 
@@ -1082,9 +1082,9 @@ The algorithm-specific fields consists of:
      the output of the encryption procedure described in
      {{ecc-kyber-encryption}}.
 
-Note that unlike ECDH, no checksum or padding are appended to the session key before key wrapping.
-Further note that unlike most public-key algorithms, in the case of a v3 PKESK packet, the symmetric algorithm identifier is not encrypted.
+Note that unlike most public-key algorithms, in the case of a v3 PKESK packet, the symmetric algorithm identifier is not encrypted.
 Instead, it is prepended to the encrypted session key in plaintext.
+In this case, the symmetric algorithm used MUST be AES-128, AES-192 or AES-256 (algorithm ID 7, 8 or 9).
 
 ### Key Material Packets {#kyber-ecc-key}
 


draft-wussler-openpgp-pqc.md
```diff
@@ -976,24 +976,28 @@ scheme is as follows:
 
 ### Public-Key Encrypted Session Key Packets (Tag 1) {#ecc-kyber-pkesk}
 
-The composite Kyber algorithms MUST be used only with v6 PKESK, as defined in
-[I-D.ietf-openpgp-crypto-refresh] Section 5.1.2.
+The algorithm-specific fields consists of:
 
-The algorithm-specific v6 PKESK parameters consists of:
+ - The one-octet algorithm identifier, if it was passed (in the case of a v3 PKESK packet).
```
wussler, line 981: I would not allow v3 PKESK with PQC. This adds IMO a security problem too: the alg ID in the v3 PKESK is inside the wrapped session key. In v6 PKESK, it is in plaintext in the encrypted data packet, but it is used again in an HKDF to derive the actual encryption key.
In this mix we open up to cross-algorithm attacks

draft-wussler-openpgp-pqc.md
```diff
@@ -976,24 +976,28 @@ scheme is as follows:
...
+ - A variable-length field containing the wrapped session key:
 
    - A one-octet size of the following field;
 
-   - Octet string of the wrapped symmetric key as described in
+   - The wrapped session key represented as an octet string, i.e.,
+     the output of the encryption procedure described in
      {{ecc-kyber-encryption}}.
 
+Note that unlike ECDH, no checksum or padding are appended to the session key before key wrapping.
```
wussler, line 997: I would not mention the differences here, this is an algorithm specific part. Let's just state how it's done here. Also AFAIK this is not done already in the new X25519 and X448 algos

draft-wussler-openpgp-pqc.md
wussler: I discussed this with Daniel and the proposed changes to the X25519/X448 algos for the refresh.
I am fine with putting the algorithm unencrypted as long as we limit the usable symmetric algorithms to AES (7, 8, 9) only. Since 7 is a must anyway, we don't limit interoperability.

This guarantees that the key size is enough to determine which algorithm ID we're using. We can still put it there explicit, but it effectively prevents cross-algorithm attacks by limiting the algorithms.

draft-wussler-openpgp-pqc.md
wussler: Yeah, but here it's not X25519 vs ECDH, here is a brand new algo. I don't know if it makes sense to add this non-normative sentence when the behavior is specified 3 lines above and the test vectors will make it clear that there is no key.

draft-wussler-openpgp-pqc.md
falko-strenzke: For the record: in the meantime we agreed that we will allow v3 PKESK for the use with PQC encryption: https://github.com/openpgp-pqc/draft-openpgp-pqc/issues/35#issuecomment-1576596480-permalink

wussler (approved):

falko-strenzke (approved):

draft-wussler-openpgp-pqc.md
falko-strenzke: I think we can mark this older conversation now as resolved (required for merging).

Issue #46, opened 2023-05-09
SHA-3 vs KMAC for KDF


teythoon:
Are there specific reasons for preferring KMAC over HKDF as key combiner?  Seeing that v6 OpenPGP now uses HKDF everywhere, it seems tempting to also use it as combiner to reduce the number of required algorithms, no?

falko-strenzke:
The problem, as far as I understand it, is that HKDF is HMAC-based and HMAC does not qualify as a dual-PRF, which is what is needed when a secret shall be derived from two input keys: https://datatracker.ietf.org/meeting/113/materials/slides-113-cfrg-a-dual-prf-construction-00#page=8

teythoon:
I don't understand.  The authors of the document are indeed using HMAC, albeit with an expansion step before, and they are doing HMAC twice, once with key 1 as key input and key 2 as data, the second time with the inputs swapped for symmetry, then xoring the HMAC outputs.

In contrast, this document simply does `KMAC(key1 || key2)`.  I don't understand why that should be a dual-PRF whereas `HKDF(key1 || key2)` should not be one.

Please unconfuse me.

The reason I'm bringing this up is that KMAC seems not to be widely available in cryptographic libraries.  A quick survey of the five cryptographic libraries that Sequoia can use paints a bleak picture:

- Nettle: not implemented
- OpenSSL: implemented in some versions, but not available in the bindings
- Botan: implementation in progress
- Rust: planned
- CNG: only in very new versions (starting with Windows 11 Preview Build 25324)

In contrast, HKDF (and HMAC) is available in all libraries in all versions we support.

falko-strenzke:
I completely understand the problem regarding the implementation aspect. Regarding the theoretical questions, I cannot address them profundly. But yes, I would also assume that an HMAC-based and thus also HKDF-based construction should be possible. It might be far from optimal in terms of efficiency though. If that doesn't matter, we could try to figure out whether this is possible.

The reason why KMAC yields a dual-PRF simply with concatenation of the keys is apparently because it is based on SHA-3. When talking about HMAC above, I refer only (and also the authors of the slides, I assume) about HMAC using SHA-2.

teythoon:
I see the statement "HMAC is generally not a dual-PRF." in the slides, but I don't see the assertion that KMAC should be one.

falko-strenzke:
[...]

PR #48, opened 2023-05-17
change KDF from KMAC to SHA3

falko-strenzke:
Some concerns regarding the availability and ease of implementation of KMAC have arisen: https://github.com/openpgp-pqc/draft-openpgp-pqc/issues/46

Accordingly, the use of readily available SHA-3 functions seems to be the superior choice.

### DIFF
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -926,7 +926,7 @@ SHA3-256 MUST be used to hash the `publicKey` of the recipient.
 For the composite KEM schemes defined in {{kem-alg-specs}} the following
 procedure MUST be used to compute the KEK that wraps a session key. The
 construction is a one-step key derivation function compliant to {{SP800-56C}}
-Section 4, based on KMAC256 {{SP800-185}}. It is given by the following
+Section 4, based on SHA3-256 {{SP800-185}}. It is given by the following
 algorithm.
 
     //   multiKeyCombine(eccKeyShare, eccCipherText,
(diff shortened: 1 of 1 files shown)


falko-strenzke:
> I do not think this is a good idea. I agree with @wussler that KMAC is the better solution. The problem that motivated the change, i.e., that the Keccak core function is not available will also manifest when implementing SHAKE which is required for Dilithium and Kyber. Hence, I do not consider it a solution to switch to SHA3 or SHA2-HMAC here.

This is not entirely true since KMAC requires cSHAKE and not SHAKE. If a library, like Botan, implements SHA-3 and SHAKE this is still not enough to implement KMAC. In such cases the Keccak[c] function has to be made available, which forces the developer to deal with its internals to some degree since the final bit padding for cSHAKE differs from that of SHAKE:

`SHAKE128(M, d) = KECCAK[256] (M || 1111, d)`

`cSHAKE128(...) = KECCAK[256](... || X || 00, L)`

This is what I experienced with Botan.

falko-strenzke:
KMAC is defined as follows:

```
KMAC256(K, X, L, S):
Validity Conditions: len(K) < 2²⁰⁴⁰ and 0 ≤ L < 2²⁰⁴⁰ and len(S) < 2²⁰⁴⁰

1. newX = bytepad(encode_string(K), 136) || X || right_encode(L).
2. T = bytepad(encode_string(“KMAC”) || encode_string(S), 136).
3. return KECCAK[512](T || newX || 00, L)  // "00" is the 2-bit padding
```

Besides some encodings it doesn't to anything different than SHA-3 hash. The necessary length encodings can easily be modelled into the KDF construction. 

Note also the virtually unlimitted key length. This is another implementation problem we experience in Botan since allowing an extremely large key length may be unexpected by the caller who might simply try to allocate `max_key_lenght`. Limiting it arbirtrarily might on the other hand limit the usability of KMAC.

I think it makes sense to hold back this PR for now and wait how the acceptance of KMAC by the community will turn out. This will probably depend on the availability of cSHAKE or KMAC in crypto libraries at that point.

PR #50, opened 2023-06-05
Added eccPublicKey to the final hashing step of the ECC-KEMs

fluppe2:
I didn't alter the security considerations, only put eccPublicKey into the hash. Shall we also elaborate on this in the security considerations?


ahuelsing (approved):

wussler:
@ahuelsing so if I understood correctly, you mean that:
- We should add the `eccPublicKey` to the hash (as done in this PR)
- We can remove the PK from the KEM combiner (in another PR)

ahuelsing:
Exactly (although I did not find a proof that guarantees that adding the eccPK to the hash does give you multi-user security, there are several papers that conjecture it does)

Issue #53, opened 2023-06-23
Ambiguity around serialization of encryption packet in section 5.2.1


alexanderkjall:
Hi

In section `5.2.1` it's stated:

```
   //   Input:
   //   algID     - the algorithm ID encoded as octet
   //   publicKey - the recipient's encryption sub-key packet
   //               serialized as octet string

   fixedInfo = algID || SHA3-256(publicKey)
```

As far as I have understood it, the sub-key packet might be represented in different ways.

This is from a conversation I had about it:

```
< teythoon> capitol: because when we hash packets or kdf them, we provide an explicit normalized form
< teythoon> capitol: normalized wrt the packet framing
< teythoon> capitol: because the ctb may be new or old, the packet length can be encoded in different ways, none of which should change the semantics
< capitol> oh, tricky
< teythoon> yeah
< teythoon> my preferred interpretation of that would be to simply exclude packet framing
``` 

Should there maybe be a clarification around this added to `5.2.1`?

wussler:
Hello @alexanderkjall, we considered to remove the PK hash in favor of hashing the PK material only with PRs #50 and #55

falko-strenzke:
The problem is solved by not using the public key as input any more.

PR #55, opened 2023-07-06
Remove PK hash from key combiner

### LINKED ISSUE #51: Remove PK hash from key combiner
Use the considerations in https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/50 and remove the duplicated PK info from the key combiner, since it's already in all KDFs

wussler:
Follow up from #50, closes #51 and #53

### DIFF
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -1559,9 +1559,15 @@ using a KEM-combiner:
 
     K = H(K1, C1, K2, C2), C = (C1, C2)
 
-We do not have to add PK1 and PK2 to H to preserve multi-user security, because
-to break security the adversary has to distinguish a set of challenge keys K*_u
-from random, one per user or public key.
+Our aim is to preserve multi-user security. A common approach to this is to add
+the public key into the key derivation for K. However, it turns out that this is
+not necessary here. To break security of the combined scheme in the multi-user
+setting, the adversary has to distinguish a set of challenge keys
+
+  K*_u = H(K1*_u, C1*_u, K2*_u, C2*_u)
+
+for users u in some set from random, also given ciphertexts `C*_u = (C1*_u,
+C2*_u)`.
 For each of these K* it holds that if the adversary never makes a query
 
     H(K1*_u, C1*_u, K2*_u, C2*_u)
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -916,10 +916,8 @@ use as binding between the KEK and the communication parties.
 
     //   Input:
     //   algID     - the algorithm ID encoded as octet
-    //   publicKey - the recipient's encryption sub-key packet
-    //               serialized as octet string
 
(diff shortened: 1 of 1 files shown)


falko-strenzke (approved):

draft-wussler-openpgp-pqc.md
```diff
@@ -1545,18 +1543,46 @@ context.
...
+
+    (c1*_u1, c2*_u1) = (c1*_u2, c2*_u2)
+
+as the ciphertext values decide for which challenge the query is meaningful.
+This means that a ciphertext collision is needed between challenges.
+Assuming that the randomness used in the generation of the two challenges is
+uncorrelated, this is negligible.
+
+In consequence, the ciphertexts already work sufficiently well as
+domain-separator.
```
wussler, line 1584: @ahuelsing I tried to convey the formal justification in https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/50 in RFC-compliant language. Please let me know if I made some mistake please, or if you think it's excessive / insufficient.

ahuelsing: Hey, I think lines 1562 ff are a bit unclear / too compact. Maybe use the following:

"Our aim is to preserve multi-user security. A common approach to this is to add the public key into the key derivation for K. However, it turns out that this is not necessary here. To break security of the combined scheme in the multi-user setting, the adversary has to distinguish a set of challenge keys 

      K*_u = H(K1*_u, C1*_u, K2*_u, C2*_u)

for users u in some set from random, also given ciphertexts   C*_u = (C1*_u, C2*_u)." 

Then continue from line 1565 on.

wussler:
```suggestion
Given two KEMs which we assume to be multi-user secure, we combine their outputs
using a KEM-combiner:

    K = H(K1, C1, K2, C2), C = (C1, C2)

Our aim is to preserve multi-user security. A common approach to this is to add the 
public key into the key derivation for K. However, it turns out that this is not necessary
here. To break security of the combined scheme in the multi-user setting, the 
adversary has to distinguish a set of challenge keys

  K*_u = H(K1*_u, C1*_u, K2*_u, C2*_u)

for users u in some set from random, also given ciphertexts C*_u = (C1*_u, C2*_u).
For each of these K* it holds that if the adversary never makes a query

    H(K1*_u, C1*_u, K2*_u, C2*_u)

they have a zero advantage over guessing.

The only multi-user advantage that the adversary could gain therefore consists
of queries to H that are meaningful for two different users u1 != u2 and their
associated public keys.
This is only the case if

    (c1*_u1, c2*_u1) = (c1*_u2, c2*_u2)

as the ciphertext values decide for which challenge the query is meaningful.
This means that a ciphertext collision is needed between challenges.
Assuming that the randomness used in the generation of the two challenges is
uncorrelated, this is negligible.

In consequence, the ciphertexts already work sufficiently well as
domain-separator.
```

Like this?

ahuelsing (approved):

Issue #57, opened 2023-08-01
Hash binding too restrictive?


falko-strenzke:
The specification of the hash binding currently doesn't allow the stronger hash functions (SHA512 vs. SHA256 and SHA3-512 vs. SHA3-256) for the smaller PQC signature parameter sets. But I think this restriction doesn't make sense. We should better both hash algorithms of the respective family for the smaller parameter set.

fluppe2:
Why should one go with SHA512 or SHA3-512 when using the Category 3 parameter sets? It shoudl be rather SHA384 or SHA3-384 for Cat. 3. But this is somehow not present in OpenPGP, is it?

However, let's discuss it.

wussler:
I think #60 addresses this well

fluppe2:
We should discuss that interactively during a meeting. I have no objections allowing more signature data digest algorithms, but I don't know if that is what is wanted in terms of simplicity.


BTW: SHA3-256 seems to be an odd choice when you aim for NIST security level 3 Dilithium, as SHA3-256 is level 2. I understand that we chose this due to lacking support for SHA3-384 in the crypto-refresh, but do we really want to go with this or ask for the inclusion if SHA3-384?

wussler:
Isn't SHA2-256 level 2? Or also SHA3?

fluppe2:
See https://csrc.nist.gov/projects/post-quantum-cryptography/post-quantum-cryptography-standardization/evaluation-criteria/security-(evaluation-criteria)

1. Any attack that breaks the relevant security definition must require computational resources comparable to or greater than those required for key search on a block cipher with a 128-bit key (e.g. AES128)

2. Any attack that breaks the relevant security definition must require computational resources comparable to or greater than those required for collision search on a 256-bit hash function (e.g. SHA256/ SHA3-256)

3. Any attack that breaks the relevant security definition must require computational resources comparable to or greater than those required for key search on a block cipher with a 192-bit key (e.g. AES192)

4. Any attack that breaks the relevant security definition must require computational resources comparable to or greater than those required for collision search on a 384-bit hash function (e.g. SHA384/ SHA3-384)

5. Any attack that breaks the relevant security definition must require computational resources comparable to or greater than those required for key search on a block cipher with a 256-bit key (e.g. AES 256)

fluppe2:
[...]

PR #60, opened 2023-08-28
Align to NIST PQC Draft Standards FIPS 203, 204, 205

fluppe2:
Take your time to review and compare with the draft standards. No need to hurry.

### The point you to some specifics changes that might be difficult to spot due to the amount of change marks:

- Felt the need to align eccKem.encap()/decap() to the ML-KEM terminology, i.e. switched to ECC-KEM.Encaps()/Decaps() (likewise x25519.Encaps() etc.)

- In Table 9 I spotted some typos (I think). The last colum was „ECDH-KEM curve“, I think it should be consistently „ECC-KEM curve“. In this last column there was X25519, X448 which should be Curve25519 and Curve448.

- In Section 5.1.1 and 5.2.4 it was:
  - "The placeholder `eccKem` has to be replaced with the specific ECC-KEM from the row "ECC-KEM" of [Table 5],  [Table 6], or [Table 7]."
  - "The placeholder `kyberKem` has to be replaced with the specific Kyber-KEM from the column "Kyber-KEM" of [Table 8]."

  I now borrowed the sentence from the NIST draft standards:

  - "To instantiate ECC-KEM, one must select a parameter set from [Table 5],  [Table 6], or [Table 7]."
  - "To instantiate ML-KEM, one must select a parameter set from the column "ML-KEM" of [Table 8]"

- In Section 5.2.4 and 5.2.5 it was:

  - "Instantiate the ECC-KEM eccKem.encap() and the ML-KEM ML-KEM.Encaps() depending on the algorithm ID according to [Table 9]"
  - "Instantiate the ECC-KEM eccKem.decap() and the ML-KEM ML-KEM.Decaps() depending on the algorithm ID according to [Table 9]"

  Felt it somehow doubled to write „instantiate ECC-KEM eccKem.encap()“. I understand that it is meant to establish the mindset „here you need encaps/decaps“. But I don’t know what it means to instantiate an algorithm and felt it better to write

  - „Instantiate the ECC-KEM and the ML-KEM depending on the algorithm ID according to [Table 9](https://github.com/openpgp-pqc/draft-openpgp-pqc/Repositories/draft-openpgp-pqc/draft-wussler-openpgp-pqc.html#tab-mlkem-ecc-composite)“

  The next lines in the procedures then say whether to Encaps or Decaps anyway.

- "Binding hashes" for ML-DSA and SLH-DSA: I switched the section title to "Signature Data Digest" to be clearer what we mean. Please take a look at the specific sections if this is ok for you as I wrote it.

### Some specific issues that I spotted and not touched but would like to pose them for discussion:

- Is it clear how to encode secret and public key material and ciphertexts/signatures from pointing the reader to the FIPS documents and saying fixed-length octet strings? Or should we list the contents of the specific octet strings or maybe point to the data in the FIPS documents?

- Section 6.2.1 states SHA3-256 as MUST and SHA3-512 as SHOULD, but if one implements ML-DSA-87 then the requirement on SHA3-512 turns into MUST. IS that clear from our formulation?

### DIFF
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -1101,7 +1101,7 @@ Algorithm ID reference | ML-DSA    | Public key | Secret key | Signature value
 
 ### Signature data digest {#mldsa-sig-data-digest}
 
-Signature data is digested prior to signing operations, see
+Signature data (i.e. the data to be signed) is digested prior to signing operations, see
 {{I-D.ietf-openpgp-crypto-refresh}} Section 5.2.4. Composite ML-DSA + ECC
 signatures MUST use the associated hash algorithm as specified in
 {{tab-mldsa-hash}} for the signature data digest. Signatures using other hash
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -778,7 +778,7 @@ Algorithm ID reference | ML-KEM                    | Public key | Secret key | C
 29, 31, 33             | ML&#x2011;KEM&#x2011;768  | 1184       | 2400       | 1088       | 32
 30, 32, 34             | ML&#x2011;KEM&#x2011;1024 | 1568       | 3168       | 1568       | 32
 
-To instantiate ML-KEM, one must select a parameter set from the column
+To instantiate `ML-KEM`, one must select a parameter set from the column
 "ML-KEM" of {{tab-mlkem-artifacts}}.
 
 The procedure to perform `ML-KEM.Encaps()` is as follows:
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -363,9 +363,10 @@ encryption schemes in order to provide a pre-quantum security fallback.
 ### ML-DSA {#mldsa-intro}
 
 ML-DSA [FIPS-204] is a signature scheme that, like ML-KEM, is based on the
-hardness of solving the Learning With Errors problem and a variant of the Short Integer Solution problem in module lattices (MLWE and SelfTargetMSIS). Accordingly, this
-specification only defines ML-DSA in composite combination with ECC-based
-signature schemes.
+hardness of solving the Learning With Errors problem and a variant of the Short
+Integer Solution problem in module lattices (MLWE and SelfTargetMSIS). 
+Accordingly, this specification only defines ML-DSA in composite combination 
+with ECC-based signature schemes.
 
 ### SLH-DSA
 
--- a/draft-wussler-openpgp-pqc.md
+++ b/draft-wussler-openpgp-pqc.md
@@ -201,31 +201,6 @@ informative:
       - org: Standards for Efficient Cryptography Group
     date: May 2009
 
-  KYBER-Subm:
-      title: CRYSTALS-Kyber (version 3.02) - Submission to round 3 of the NIST post-quantum project
-      author:
-        -
-          ins: R. Avanzi
-        -
-          ins: J. Bos
-        -
-          ins: L. Ducas
-        -
-          ins: E. Kiltz
-        -
-          ins: T. Lepoint
-        -
-          ins: V. Lyubashevsky
-        -
-          ins: J. M. Schanck
-        -
-          ins: P. Schwabe
-        -
-          ins: G. Seiler
-        -
-          ins: D. Stehle
-      date: 2021-08-04
-
   FIPS-203:
       target: https://doi.org/10.6028/NIST.FIPS.203.ipd
       title: Module-Lattice-Based Key-Encapsulation Mechanism Standard
@@ -644,7 +619,7 @@ for future extensions.
 
 ## Composite KEMs
 
-Kyber + ECC public-key encryption is meant to involve both the Kyber KEM and an
+ML-KEM + ECC public-key encryption is meant to involve both the ML-KEM and an
 ECC-based KEM in an a priori non-separable manner. This is achieved via KEM
 combination, i.e. both key encapsulations/decapsulations are performed in
 parallel, and the resulting key shares are fed into a key combiner to produce a
@@ -660,7 +635,7 @@ recipient.
 
 ## Composite Signatures
 
-Dilithium + ECC signatures are meant to contain both the Dilithium and the ECC
+ML-DSA + ECC signatures are meant to contain both the ML-DSA and the ECC
 signature data, and an implementation MUST validate both algorithms to state
 that a signature is valid.
 
@@ -744,14 +719,14 @@ The various procedures to perform the operations of an ECC-based KEM are
 defined in the following subsections. Specifically, each of these subsections
 defines the instances of the following operations:
 
-    (eccCipherText, eccKeyShare) <- eccKem.encap(eccPublicKey)
+    (eccCipherText, eccKeyShare) <- ECC-KEM.Encaps(eccPublicKey)
 
 and
 
-    (eccKeyShare) <- eccKem.decap(eccPrivateKey, eccCipherText)
+    (eccKeyShare) <- ECC-KEM.Decaps(eccPrivateKey, eccCipherText)
 
-The placeholder `eccKem` has to be replaced with the specific ECC-KEM from the
-row "ECC-KEM" of {{tab-ecdh-cfrg-artifacts}}, {{tab-ecdh-nist-artifacts}}, and
+To instantiate ECC-KEM, one must select a parameter set from
+{{tab-ecdh-cfrg-artifacts}}, {{tab-ecdh-nist-artifacts}}, or
 {{tab-ecdh-brainpool-artifacts}}.
 
 #### X25519-KEM {#x25519-kem}
@@ -762,7 +737,7 @@ using the function `X25519()` and encodings defined in [RFC7748]. The
 to the equation `R = X25519(r, U(P))`. Here, `U(P)` denotes the u-coordinate of
 the base point of Curve25519.
 
-The operation `x25519Kem.encap()` is defined as follows:
+The operation `x25519Kem.Encaps()` is defined as follows:
 
  1. Generate an ephemeral key pair {`v`, `V`} via `V = X25519(v,U(P))` where `v`
     is a random scalar
@@ -774,7 +749,7 @@ The operation `x25519Kem.encap()` is defined as follows:
 
  4. Set the output `eccKeyShare` to `SHA3-256(X || eccCipherText || eccPublicKey)`
 
-The operation `x25519Kem.decap()` is defined as follows:
+The operation `x25519Kem.Decaps()` is defined as follows:
 
  1. Compute the shared coordinate `X = X25519(r, V)`, where `r` is the
     `eccPrivateKey` and `V` is the `eccCipherText`
@@ -789,7 +764,7 @@ is denoted as `r`, the `eccPublicKey` as `R`, they are subject to the equation
 `R = X25519(r, U(P))`. Here, `U(P)` denotes the u-coordinate of the base point
 of Curve448.
 
-The operation `x448.encap()` is defined as follows:
+The operation `x448.Encaps()` is defined as follows:
 
  1. Generate an ephemeral key pair {`v`, `V`} via `V = X448(v,U(P))` where `v`
     is a random scalar
@@ -801,7 +776,7 @@ The operation `x448.encap()` is defined as follows:
 
  4. Set the output `eccKeyShare` to `SHA3-512(X || eccCipherText || eccPublicKey)`
 
-The operation `x448Kem.decap()` is defined as follows:
+The operation `x448Kem.Decaps()` is defined as follows:
 
  1. Compute the shared coordinate `X = X448(r, V)`, where `r` is the
     `eccPrivateKey` and `V` is the `eccCipherText`
@@ -810,7 +785,7 @@ The operation `x448Kem.decap()` is defined as follows:
 
 #### ECDH-KEM {#ecdh-kem}
 
-The operation `ecdhKem.encap()` is defined as follows:
+The operation `ecdhKem.Encaps()` is defined as follows:
 
  1. Generate an ephemeral key pair {`v`, `V=vG`} as defined in
     {{SP800-186}} or {{RFC5639}} where `v` is a random scalar
@@ -827,7 +802,7 @@ The operation `ecdhKem.encap()` is defined as follows:
     chosen according to {{tab-ecdh-nist-artifacts}} or
     {{tab-ecdh-brainpool-artifacts}}
 
-The operation `ecdhKem.decap()` is defined as follows:
+The operation `ecdhKem.Decaps()` is defined as follows:
 
  1. Compute the shared Point `S` as `rV`, where `r` is the `eccPrivateKey` and
     `V` is the `eccCipherText`, according to {{SP800-186}} or {{RFC5639}}
@@ -839,69 +814,70 @@ The operation `ecdhKem.decap()` is defined as follows:
     chosen according to {{tab-ecdh-nist-artifacts}} or
     {{tab-ecdh-brainpool-artifacts}}
 
-### Kyber-KEM {#kyber-kem}
+### ML-KEM {#mlkem-ops}
 
-Kyber-KEM features the following operations:
+ML-KEM features the following operations:
 
-    (kyberCipherText, kyberKeyShare) <- kyberKem.encap(kyberPublicKey)
+    (mlkemCipherText, mlkemKeyShare) <- ML-KEM.Encaps(mlkemEncapsKey)
 
 and
 
-    (kyberKeyShare) <- kyberKem.decap(kyberCipherText, kyberPrivateKey)
+    (mlkemKeyShare) <- ML-KEM.Decaps(mlkemCipherText, mlkemDecapsKey)
 
-The above are the operations Kyber.CCAKEM.Enc() and Kyber.CCAKEM.Dec() defined
-in [Kyber-Subm].
+The above are the operations ML-KEM.Encaps() and ML-KEM.Decaps() defined in
+[FIPS-203]. Note that `mlkemEncapsKey` is public and `mlkemDecapsKey` is
+private keying material.
 
-Kyber-KEM has the parameterization with the corresponding artifact lengths in
-octets as given in {{tab-kyber-artifacts}}. All artifacts are encoded as
-defined in [Kyber-Subm].
+ML-KEM has the parameterization with the corresponding artifact lengths in
+octets as given in {{tab-mlkem-artifacts}}. All artifacts are encoded as
+defined in [FIPS-203].
 
-{: title="Kyber-KEM parameters artifact lengths in octets" #tab-kyber-artifacts}
-Algorithm ID reference | Kyber-KEM    | Public key | Secret key | Ciphertext | Key share
+{: title="ML-KEM parameters artifact lengths in octets" #tab-mlkem-artifacts}
+Algorithm ID reference | ML-KEM       | Public key | Secret key | Ciphertext | Key share
 ----------------------:| ------------ | ---------- | ---------- | ---------- | ---------
-29, 31, 33             | kyberKem768  | 1184       | 2400       | 1088       | 32
-30, 32, 34             | kyberKem1024 | 1568       | 3186       | 1568       | 32
+29, 31, 33             | ML-KEM-768   | 1184       | 2400       | 1088       | 32
+30, 32, 34             | ML-KEM-1024  | 1568       | 3168       | 1568       | 32
 
-The placeholder `kyberKem` has to be replaced with the specific Kyber-KEM from
-the column "Kyber-KEM" of {{tab-kyber-artifacts}}.
+To instantiate ML-KEM, one must select a parameter set from the column
+"ML-KEM" of {{tab-mlkem-artifacts}}.
 
-The procedure to perform `kyberKem.encap()` is as follows:
+The procedure to perform `ML-KEM.Encaps()` is as follows:
 
- 1. Extract the component public key `kyberPublicKey` that is part of the
+ 1. Extract the encapsulation key `mlkemEncapsKey` that is part of the
     recipient's composite public key
 
- 2. Invoke `(kyberCipherText, keyShare) <- kyberKem.encap(kyberPublicKey)`
+ 2. Invoke `(mlkemCipherText, mlkemKeyShare) <- ML-KEM.Encaps(mlkemEncapsKey)`
 
- 3. Set `kyberCipherText` as the Kyber ciphertext
+ 3. Set `mlkemCipherText` as the ML-KEM ciphertext
 
- 4. Set `keyShare` as the Kyber symmetric key share
+ 4. Set `mlkemKeyShare` as the ML-KEM symmetric key share
 
-The procedure to perform `kyberKem.decap()` is as follows:
+The procedure to perform `ML-KEM.Decaps()` is as follows:
 
- 1. Invoke `keyShare <-  kyberKem.decap(kyberCipherText, kyberPrivateKey)`
+ 1. Invoke `mlkemKeyShare <-  ML-KEM.Decaps(mlkemCipherText, mlkemDecapsKey)`
 
- 2. Set `keyShare` as the Kyber symmetric key
+ 2. Set `mlkemKeyShare` as the ML-KEM symmetric key share
 
-## Composite Encryption Schemes with Kyber {#ecc-mlkem}
+## Composite Encryption Schemes with ML-KEM {#ecc-mlkem}
 
-{{kem-alg-specs}} specifies the following Kyber + ECC composite public-key
+{{kem-alg-specs}} specifies the following ML-KEM + ECC composite public-key
 encryption schemes:
 
-{: title="Kyber-ECC-composite Schemes" #tab-kyber-ecc-composite}
-Algorithm ID reference | Kyber-KEM    | ECC-KEM   | ECDH-KEM curve
+{: title="ML-KEM + ECC composite schemes" #tab-mlkem-ecc-composite}
+Algorithm ID reference | ML-KEM       | ECC-KEM   | ECC-KEM curve
 ----------------------:| ------------ | --------- | --------------
-29                     | kyberKem768  | x25519Kem | X25519
-30                     | kyberKem1024 | x448Kem   | X448
-31                     | kyberKem768  | ecdhKem   | NIST P-256
-32                     | kyberKem1024 | ecdhKem   | NIST P-384
-33                     | kyberKem768  | ecdhKem   | brainpoolP256r1
-34                     | kyberKem1024 | ecdhKem   | brainpoolP384r1
+29                     | ML-KEM-768   | x25519Kem | Curve25519
+30                     | ML-KEM-1024  | x448Kem   | Curve448
+31                     | ML-KEM-768   | ecdhKem   | NIST P-256
+32                     | ML-KEM-1024  | ecdhKem   | NIST P-384
+33                     | ML-KEM-768   | ecdhKem   | brainpoolP256r1
+34                     | ML-KEM-1024  | ecdhKem   | brainpoolP384r1
 
-The Kyber + ECC composite public-key encryption schemes are built according to
+The ML-KEM + ECC composite public-key encryption schemes are built according to
 the following principal design:
 
- - The Kyber-KEM encapsulation algorithm is invoked to create a Kyber
-   ciphertext together with a Kyber symmetric key share.
+ - The ML-KEM encapsulation algorithm is invoked to create a ML-KEM ciphertext
+   together with a ML-KEM symmetric key share.
 
  - The encapsulation algorithm of an ECC-based KEM, namely one out of
    X25519-KEM, X448-KEM, or ECDH-KEM is invoked to create an ECC ciphertext
@@ -914,7 +890,7 @@ the following principal design:
  - The session key for content encryption is then wrapped as described in
    {{RFC3394}} using AES-256 as algorithm and the KEK as key.
 
- - The PKESK package's algorithm-specific parts are made up of the Kyber
+ - The PKESK package's algorithm-specific parts are made up of the ML-KEM
    ciphertext, the ECC ciphertext, and the wrapped session key.
 
 ### Fixed information {#kem-fixed-info}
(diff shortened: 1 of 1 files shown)


draft-wussler-openpgp-pqc.md
```diff
@@ -1057,47 +979,48 @@ scheme is as follows:
...
+   - The wrapped session key represented as an octet string, i.e., the output
+     of the encryption procedure described in {{ecc-mlkem-encryption}}.
 
-Note that unlike most public-key algorithms, in the case of a v3 PKESK packet, the symmetric algorithm identifier is not encrypted.
-Instead, it is prepended to the encrypted session key in plaintext.
-In this case, the symmetric algorithm used MUST be AES-128, AES-192 or AES-256 (algorithm ID 7, 8 or 9).
+Note that unlike most public-key algorithms, in the case of a v3 PKESK packet,
+the symmetric algorithm identifier is not encrypted.  Instead, it is prepended
+to the encrypted session key in plaintext.  In this case, the symmetric
+algorithm used MUST be AES-128, AES-192 or AES-256 (algorithm ID 7, 8 or 9).
```
fluppe2, line 1005:
```suggestion
algorithm used MUST be AES-128, AES-192 or AES-256.
```
Is it really necessary to list the algorithm IDs for AES from another spec?

draft-wussler-openpgp-pqc.md
```diff
@@ -401,47 +338,33 @@ abbreviation "PQ/T" is used. The short form "PQ(/T)" stands for PQ or PQ/T.
...
+ML-KEM [FIPS-203] is based on the hardness of solving the learning-with-errors
+problem in module lattices (MLWE). The scheme is believed to provide security
+against cryptanalytic attacks by classical as well as quantum computers. This
+specification defines ML-KEM only in composite combination with ECC-based
+encryption schemes in order to provide a pre-quantum security fallback.
+
+### ML-DSA {#mldsa-intro}
+
+ML-DSA [FIPS-204] is a signature scheme that, like ML-KEM, is based on the
+hardness of solving lattice problems in module lattices. Accordingly, this
```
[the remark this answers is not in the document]
fluppe2: I think the reason why we phrased it differently is because the security of ML-DSA bases on different problems (MLWE and MSIS). See §3.2 of FIPS-204. I guess there is space for optimization.

draft-wussler-openpgp-pqc.md
```diff
@@ -946,13 +871,13 @@ algorithm.
     //   customizationString - the UTF-8 encoding of the string "KDF"
 
     eccKemData = eccKeyShare || eccCipherText
-    kyberKemData = kyberKeyShare || kyberCipherText
-    encData = counter || eccKemData || kyberKemData || fixedInfo
+    mlkemKemData = mlkemKeyShare || mlkemCipherText
```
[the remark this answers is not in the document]
fluppe2: Good point, I opted for the first suggestion.

draft-wussler-openpgp-pqc.md
```diff
@@ -365,21 +302,21 @@ The algorithms defined in this extension were chosen for standardization by the
...
+CRYSTALS-Dilithium) as well as SLH-DSA (formerly SPHINCS+) as signature
+schemes.
 
-For the two CRYSTALS-* schemes, this document follows the conservative strategy
-to deploy post-quantum in combination with traditional schemes such that the
+For the two ML-* schemes, this document follows the conservative strategy to
+deploy post-quantum in combination with traditional schemes such that the
 security is retained even if all schemes but one in the combination are broken.
-In contrast, the hashed-based signature scheme SPHINCS+ is considered to be
+In contrast, the hashed-based signature scheme SLH-DSA is considered to be
```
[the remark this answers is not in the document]
fluppe2: Good catch

draft-wussler-openpgp-pqc.md
```diff
@@ -1028,27 +952,25 @@ scheme is as follows:
 
  3. Check that the own and the extracted algorithm ID match
 
- 4. Parse the `eccSecretKey` and `kyberSecretKey` from the algorithm specific
+ 4. Parse the `eccSecretKey` and `mlkemDecapsKey` from the algorithm specific
```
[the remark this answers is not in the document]
fluppe2: Good catch. Changed `eccSecretKey` to `eccPrivateKey`

And yes, I adopted the terminology from ML-KEM specification. This is a remark I forgot to write in the PR description above. We should discuss that. I introduced this terminology in §5.1.2, saying that the encapsulation key is public and the decapsulation key is private keying material. I think since we have composite public and private keys which are composed of the ECC private/public and the ML-KEM Decapsulation/Encapsulation keys it does no harm to adopt the terminology of NIST.

draft-wussler-openpgp-pqc.md
```diff
@@ -1281,114 +1197,105 @@ consists of:
...
-3                      | 192s                  | 48                  | 96                  | 16224
-4                      | 192f                  | 48                  | 96                  | 35664
-5                      | 256s                  | 64                  | 128                 | 29792
-6                      | 256f                  | 64                  | 128                 | 49856
+### Signature Data Digest
 
-### Binding hashes
+SLH-DSA signatures MUST use the associated hash as specified in
+{{tab-slhdsa-hash}} for the digest of signature data according to
+{{I-D.ietf-openpgp-crypto-refresh}} Section 5.2.4. Signatures using other hash
```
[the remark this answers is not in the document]
fluppe2: I just turned the sentence around. I want to say that the signature data digest before the signature algorithm must be done via a hash algorithm from the table. Aron called this "binding hashes".

fluppe2:
> * One occurance of Dilithium is still here: "far greater than those of traditional or Dilithium + ECC signature schemes" (Section 10.1.)

Good catch, aligned in commit [6095c0f](https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60/commits/6095c0f84f52a68f60f637c592545d909eccedf4).

draft-wussler-openpgp-pqc.md
```diff
@@ -1281,114 +1197,105 @@ consists of:
...
-1                      | 128s                  | 32                  | 64                  | 7856
-2                      | 128f                  | 32                  | 64                  | 17088
-3                      | 192s                  | 48                  | 96                  | 16224
-4                      | 192f                  | 48                  | 96                  | 35664
-5                      | 256s                  | 64                  | 128                 | 29792
-6                      | 256f                  | 64                  | 128                 | 49856
+### Signature Data Digest
 
-### Binding hashes
+SLH-DSA signatures MUST use the associated hash as specified in
```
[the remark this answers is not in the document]
fluppe2: See [6095c0f](https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60/commits/6095c0f84f52a68f60f637c592545d909eccedf4).

draft-wussler-openpgp-pqc.md
```diff
@@ -1150,129 +1073,122 @@ Algorithm ID reference | Curve           | Field size | Public key | Secret key
...
 
-{: title="Dilithium parameters and artifact lengths in octets" #tab-dilithium-artifacts}
-Algorithm ID reference | Dilithium instance | Public key | Secret key | Signature value
-----------------------:| ------------------ | -----------| ---------- | ---------------
-35, 37, 39             | Dilithium3         | 1952       | 4000       | 3293
-36, 38, 40             | Dilithium5         | 2592       | 4864       | 4595
+{: title="ML-DSA parameters and artifact lengths in octets" #tab-mldsa-artifacts}
+Algorithm ID reference | ML-DSA    | Public key | Secret key | Signature value
+----------------------:| --------- | -----------| ---------- | ---------------
+35, 37, 39             | ML-DSA-65 | 1952       | 4000       | 3293
```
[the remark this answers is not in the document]
fluppe2: See [f5aeef2](https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60/commits/f5aeef293a70dd6cae838925b71e82e4f6e5e5b9).

draft-wussler-openpgp-pqc.md
```diff
@@ -830,69 +754,70 @@ The operation `ecdhKem.decap()` is defined as follows:
...
 
-{: title="Kyber-KEM parameters artifact lengths in octets" #tab-kyber-artifacts}
-Algorithm ID reference | Kyber-KEM    | Public key | Secret key | Ciphertext | Key share
+{: title="ML-KEM parameters artifact lengths in octets" #tab-mlkem-artifacts}
+Algorithm ID reference | ML-KEM       | Public key | Secret key | Ciphertext | Key share
 ----------------------:| ------------ | ---------- | ---------- | ---------- | ---------
-29, 31, 33             | kyberKem768  | 1184       | 2400       | 1088       | 32
-30, 32, 34             | kyberKem1024 | 1568       | 3186       | 1568       | 32
+29, 31, 33             | ML-KEM-768   | 1184       | 2400       | 1088       | 32
+30, 32, 34             | ML-KEM-1024  | 1568       | 3168       | 1568       | 32
```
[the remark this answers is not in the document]
fluppe2: See [f5aeef2](https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60/commits/f5aeef293a70dd6cae838925b71e82e4f6e5e5b9).

fluppe2:
> For me it's clear but I see what you mean. Perhaps it makes sense to replace the text by a list that binds each algorithm ID to a hash algorithm, analogous to the SLH-DSA parameters (Table 14).
> We can also address #57 there (i.e., SHA3-512 is always allowed when SHA3-256 is allowed).

See [16f5b8b](https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60/commits/16f5b8bdf922f468bef6829518cab4c31bdbad32)

I guess this closes #57

draft-wussler-openpgp-pqc.md
wussler: I don't think it's necessary, but makes the dev life slightly easier IMO

draft-wussler-openpgp-pqc.md
```diff
@@ -735,14 +659,14 @@ The various procedures to perform the operations of an ECC-based KEM are
...
+    (eccCipherText, eccKeyShare) <- ECC-KEM.Encaps(eccPublicKey)
 
 and
 
-    (eccKeyShare) <- eccKem.decap(eccPrivateKey, eccCipherText)
+    (eccKeyShare) <- ECC-KEM.Decaps(eccPrivateKey, eccCipherText)
 
-The placeholder `eccKem` has to be replaced with the specific ECC-KEM from the
-row "ECC-KEM" of {{tab-ecdh-cfrg-artifacts}}, {{tab-ecdh-nist-artifacts}}, and
+To instantiate ECC-KEM, one must select a parameter set from
```
wussler, line 668:
```suggestion
To instantiate `ECC-KEM`, one must select a parameter set from
```

draft-wussler-openpgp-pqc.md
```diff
@@ -830,69 +754,70 @@ The operation `ecdhKem.decap()` is defined as follows:
...
-{: title="Kyber-KEM parameters artifact lengths in octets" #tab-kyber-artifacts}
-Algorithm ID reference | Kyber-KEM    | Public key | Secret key | Ciphertext | Key share
-----------------------:| ------------ | ---------- | ---------- | ---------- | ---------
-29, 31, 33             | kyberKem768  | 1184       | 2400       | 1088       | 32
-30, 32, 34             | kyberKem1024 | 1568       | 3186       | 1568       | 32
+{: title="ML-KEM parameters artifact lengths in octets" #tab-mlkem-artifacts}
+Algorithm ID reference | ML-KEM                    | Public key | Secret key | Ciphertext | Key share
+----------------------:| ------------------------- | ---------- | ---------- | ---------- | ---------
+29, 31, 33             | ML&#x2011;KEM&#x2011;768  | 1184       | 2400       | 1088       | 32
+30, 32, 34             | ML&#x2011;KEM&#x2011;1024 | 1568       | 3168       | 1568       | 32
```
wussler, line 779:
```suggestion
Algorithm ID reference | ML-KEM      | Public key | Secret key | Ciphertext | Key share
----------------------:| ----------- | ---------- | ---------- | ---------- | ---------
29, 31, 33             | ML-KEM-768  | 1184       | 2400       | 1088       | 32
30, 32, 34             | ML-KEM-1024 | 1568       | 3168       | 1568       | 32
```

draft-wussler-openpgp-pqc.md
```diff
@@ -830,69 +754,70 @@ The operation `ecdhKem.decap()` is defined as follows:
...
-29                     | kyberKem768  | x25519Kem | X25519
-30                     | kyberKem1024 | x448Kem   | X448
-31                     | kyberKem768  | ecdhKem   | NIST P-256
-32                     | kyberKem1024 | ecdhKem   | NIST P-384
-33                     | kyberKem768  | ecdhKem   | brainpoolP256r1
-34                     | kyberKem1024 | ecdhKem   | brainpoolP384r1
-
-The Kyber + ECC composite public-key encryption schemes are built according to
+29                     | ML-KEM-768   | x25519Kem | Curve25519
+30                     | ML-KEM-1024  | x448Kem   | Curve448
```
wussler, line 810: Why changing `X*` to `Curve*`?

draft-wussler-openpgp-pqc.md
```diff
@@ -1150,129 +1073,128 @@ Algorithm ID reference | Curve           | Field size | Public key | Secret key
...
-{: title="Dilithium parameters and artifact lengths in octets" #tab-dilithium-artifacts}
-Algorithm ID reference | Dilithium instance | Public key | Secret key | Signature value
-----------------------:| ------------------ | -----------| ---------- | ---------------
-35, 37, 39             | Dilithium3         | 1952       | 4000       | 3293
-36, 38, 40             | Dilithium5         | 2592       | 4864       | 4595
+{: title="ML-DSA parameters and artifact lengths in octets" #tab-mldsa-artifacts}
+Algorithm ID reference | ML-DSA                  | Public key | Secret key | Signature value
+----------------------:| ----------------------- | -----------| ---------- | ---------------
+35, 37, 39             | ML&#x2011;DSA&#x2011;65 | 1952       | 4000       | 3293
+36, 38, 40             | ML&#x2011;DSA&#x2011;87 | 2592       | 4864       | 4595
```
wussler, line 1098:
```suggestion
Algorithm ID reference | ML-DSA    | Public key | Secret key | Signature value
----------------------:| --------- | -----------| ---------- | ---------------
35, 37, 39             | ML-DSA-65 | 1952       | 4000       | 3293
36, 38, 40             | ML-DSA-87 | 2592       | 4864       | 4595
```

draft-wussler-openpgp-pqc.md
fluppe2: So keep it as it was (listing the IDs)?
It's ok for me.

draft-wussler-openpgp-pqc.md
fluppe2: Because the column lists the curves associated to the algorithm. X* denotes algorithms/functions and Curve* denotes the curves. I felt it should be consistent.

draft-wussler-openpgp-pqc.md
fluppe2: @TJ-91 suggested to insert those hyphen characters in https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60#discussion_r1307226240 in order to avoid unwanted line breaks. Is it not breaking weirdly on your side?

draft-wussler-openpgp-pqc.md
fluppe2: See https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60#discussion_r1311121166

draft-wussler-openpgp-pqc.md
wussler: Yes, as much as we list the IDs in the hash binding tables

draft-wussler-openpgp-pqc.md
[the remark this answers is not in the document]
wussler: Hmm, I see, but this breaks the plaintext version
image

draft-wussler-openpgp-pqc.md
fluppe2: Ok, then lets drop this suggestion.

draft-wussler-openpgp-pqc.md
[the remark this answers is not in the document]
fluppe2: @sehlen Are you fine with this formulation? FIPS-204 states that security goes back to a non-standard version of MSIS.

draft-wussler-openpgp-pqc.md
fluppe2: Applied inline code marking also to ML-KEM in the anlagous sentence. See https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60/commits/fa1b1284d6bea9ceb1dafbbf470305744c0f0fe7

draft-wussler-openpgp-pqc.md
fluppe2: I opted for Aron's suggestion to avoid any kind of special character encodings. I think that ends up messy. Let it break as it breaks.

draft-wussler-openpgp-pqc.md
fluppe2: Let it break as it break, no special character encodings.

falko-strenzke:
> Felt it somehow doubled to write „instantiate ECC-KEM eccKem.encap()“. I understand that it is meant to establish the mindset „here you need encaps/decaps“. But I don’t know what it means to instantiate an algorithm and felt it better to write
> 
>     * „Instantiate the ECC-KEM and the ML-KEM depending on the algorithm ID according to [Table 9](https://github.com/openpgp-pqc/draft-openpgp-pqc/Repositories/draft-openpgp-pqc/draft-wussler-openpgp-pqc.html#tab-mlkem-ecc-composite)“

The "instantiate" terminology is in my understanding indeed a somewhat object oriented concept for the description of the algorithms (in an abstract sense, not with reference to the programming approach). It basically means: "select the set of algorithms / parameters according to the respective KEM (or signature) variant. I still think it can be a valid and intuitively understandable verbal approach if we use it consistently. But I am aware that it is not necessarily the way that specs are typically written and I am also OK with removing the term "instantiate" completely.

draft-wussler-openpgp-pqc.md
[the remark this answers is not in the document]
falko-strenzke: Is "signature data" really the correct term for the data that is signed? To me it seems it could also be understood as the data that represents the signature. Maybe we should better call it "to-be-signed data"?

draft-wussler-openpgp-pqc.md
```diff
@@ -401,47 +338,33 @@ abbreviation "PQ/T" is used. The short form "PQ(/T)" stands for PQ or PQ/T.
 This section describes the individual post-quantum cryptographic schemes. All
 schemes listed here are believed to provide security in the presence of a
 cryptographically relevant quantum computer. However, the mathematical problems
-on which the two CRYSTALS-* schemes and SPHINCS+ are based, are fundamentally
+on which the two ML-* schemes and SLH-DSA are based, are fundamentally
 different, and accordingly the level of trust commonly placed in them as well
 as their performance characteristics vary.
 
-\[Note to the reader: This specification refers to the latest NIST submission
```
falko-strenzke, line 408: I wouldn't completely remove this editorial note yet, but only adapt it to the new situation that now we have NIST drafts for the algorithms. But it is still relevant to inform readers that these are only drafts and still subject to change.

draft-wussler-openpgp-pqc.md
```diff
@@ -1057,47 +979,48 @@ scheme is as follows:
...
+### Key Material Packets {#mlkem-ecc-key}
 
 The algorithm-specific public key is this series of values:
 
  - A fixed-length octet string representing an EC point public key, in the
    point format associated with the curve specified in {{ecc-kem}}.
 
- - A fixed-length octet string containing the Kyber public key, whose length
-   depends on the algorithm ID as specified in {{tab-kyber-artifacts}}.
+ - A fixed-length octet string containing the ML-KEM encapsulation key, whose
```
falko-strenzke, line 1014: I am not sure about the change to "encapsulation/decapsulation key" from "public/private key". Otherwise, in order to be consistent, we would have to eliminate all uses of "public/private key" and replace it with "encryption/decryption key" or "verification/signature key". Does anyone see a specific reason why KEM keys should be handled differently in this respect?

draft-wussler-openpgp-pqc.md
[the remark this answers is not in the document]
falko-strenzke: As @fluppe2 pointed out, the above use of the term "signature data" is consistent with the [crypto refresh](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-10.html#section-5.2.4).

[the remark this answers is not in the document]
falko-strenzke: I suggest to add a clarifying subclause "for the digest of the signature data (data to be signed)"

draft-wussler-openpgp-pqc.md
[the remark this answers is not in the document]
falko-strenzke: team agreed on this suggestion on 2023-09-11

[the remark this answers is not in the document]
fluppe2: Team agrees on:
- eccPublicKey and eccSecretKey
- mlkemPublicKey and mlkemSecretKey
- ecdsaPublicKey and ecdsaSecretKey
- eddsaPublicKey and eddsaSecretKey

Todo @fluppe2

draft-wussler-openpgp-pqc.md
```diff
@@ -1150,129 +1073,128 @@ Algorithm ID reference | Curve           | Field size | Public key | Secret key
...
-## Composite Signature Schemes with Dilithium {#ecc-dilithium}
+## Composite Signature Schemes with ML-DSA {#ecc-mldsa}
 
-### Binding hashes
+### Signature data digest {#mldsa-sig-data-digest}
 
-Composite Dilithium + ECC signatures MUST use SHA3-256 (hash algorithm ID 12)
-or SHA3-512 (hash algorithm ID 14) as hashing algorithm. Signatures using other
-hash algorithms MUST be considered invalid.
+Signature data is digested prior to signing operations, see
```
fluppe2, line 1104:
```suggestion
Signature data (i.e. the data to be signed) is digested prior to signing operations, see
```

draft-wussler-openpgp-pqc.md
```diff
@@ -1281,114 +1203,106 @@ consists of:
...
-1                      | 128s                  | 32                  | 64                  | 7856
-2                      | 128f                  | 32                  | 64                  | 17088
-3                      | 192s                  | 48                  | 96                  | 16224
-4                      | 192f                  | 48                  | 96                  | 35664
-5                      | 256s                  | 64                  | 128                 | 29792
-6                      | 256f                  | 64                  | 128                 | 49856
+### Signature Data Digest {#slhdsa-sig-data-digest}
 
-### Binding hashes
+Signature data is digested prior to signing operations, see
```
fluppe2, line 1270:
```suggestion
Signature data (i.e. the data to be signed) is digested prior to signing operations, see
```
As discussed in team on 2023-09-11.

draft-wussler-openpgp-pqc.md
fluppe2: Team decision: Keep the note but switch from "NIST submission" to "NIST draft standards".

Todo @falko-strenzke

falko-strenzke: This is my proposal for the updated note:
``` 
\[Note to the reader: This specification refers to the NIST PQC standards drafts
Draft FIPS 203, Draft FIPS 204, and Draft FIPS 205 as if it were a final
specification. This is a temporary solution until the final versions of these
documents are available. The goal is to provide a sufficiently precise
specification of the algorithms already at the draft stage of this
specification, so that it is possible for implementers to create interoperable
implementations. Furthermore, we want to point out that, depending on possible
future changes to the draft standards by NIST, this specification may be updated
as soon as corresponding information becomes available.\]
```
Maybe we should indicate that we are referring to NIST **Drafts** also in the references by prefixing of postfixing the title either with "FIPS 203 (Initial Public Draft)" etc. as it is on the NIST web page of "FIPS 203 (Draft)" as it is in the PDF of the draft.

@fluppe2 Do you want to include the new version of the note into the PR (possibly after revising it if necessary from your point of view)? This is maybe easier than creating a separate PR or me working in this PR.

fluppe2: Done in https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60/commits/6193c31b7ac32c11b9183d77ce7721f15aeae478

fluppe2:
I guess we are good to go, once the approving reviews come together.

fluppe2:
To avoid rebasing I adressed #62 in https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60/commits/ffbf99672294d5920a02db58b5e9adadbc8fa113. I guess we can close #62.

fluppe2:
> We still have five occurances of "private key". Should we rename them to "secret key" as well?

Fixed in https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/60/commits/e69fe9de058cd9283ba2e9a4e8a0df1d154b8b1f

draft-wussler-openpgp-pqc.md
```diff
@@ -1281,114 +1213,106 @@ consists of:
...
+### Signature Data Digest {#slhdsa-sig-data-digest}
 
-SPHINCS+ signature packets MUST use the associated hash as specified in
-{{tab-sphincs-hash}}. Signature packets using other hashes MUST be considered
-invalid.
+Signature data (i.e. the data to be signed) is digested prior to signing operations, see
+{{I-D.ietf-openpgp-crypto-refresh}} Section 5.2.4. SLH-DSA signatures MUST use
+the associated hash algorithm as specified in {{tab-slhdsa-hash}} for the
+signature data digest. Signatures using other hash algorithms MUST be
+considered invalid.
```
wussler, line 1284:
```suggestion
Signature data (i.e. the data to be signed) is digested prior to signing
operations, see {{I-D.ietf-openpgp-crypto-refresh}} Section 5.2.4. SLH-DSA
signatures MUST use the associated hash algorithm as specified in
{{tab-slhdsa-hash}} for the signature data digest. Signatures using other hash
algorithms MUST be considered invalid.
```

draft-wussler-openpgp-pqc.md
```diff
@@ -365,21 +302,21 @@ The algorithms defined in this extension were chosen for standardization by the
...
+deploy post-quantum in combination with traditional schemes such that the
 security is retained even if all schemes but one in the combination are broken.
-In contrast, the hashed-based signature scheme SPHINCS+ is considered to be
-sufficiently well understood with respect to its security assumptions in order
-to be used standalone. To this end, this document specifies the following new
-set: SPHINCS+ standalone and CRYSTALS-* as composite with ECC-based KEM and
+In contrast, the stateless hash-based signature scheme SLH-DSA is considered to
+be sufficiently well understood with respect to its security assumptions in
+order to be used standalone. To this end, this document specifies the following
+new set: SLH-DSA standalone and ML-* as composite with ECC-based KEM and
```
falko-strenzke, line 316:
```suggestion
new set: SLH-DSA standalone and two ML-* as composite with ECC-based KEM and
```

Using regular expression like syntax in normal text is not necessarily common or intuitive to readers. Adding the word "two" makes it at least clearer that this is refering to two schemes and not denoting a specific one.

draft-wussler-openpgp-pqc.md
```diff
@@ -1565,84 +1489,88 @@ setting, the adversary has to distinguish a set of challenge keys
...
-This specification introduces both Dilithium + ECC as well as SPHINCS+ as
-PQ(/T) signature schemes.
+This specification introduces both ML-DSA + ECC as well as SLH-DSA as PQ(/T)
+signature schemes.
 
-Generally, it can be said that Dilithium + ECC provides a performance in terms
-of execution time and space requirements that is close to that of traditional
-ECC signature schemes. Implementers may want to offer SPHINCS+ for applications
+Generally, it can be said that ML-DSA + ECC provides a performance in terms of
+execution time and space requirements that is close to that of traditional ECC
```
falko-strenzke, line 1541: Hmm, I think regarding execution time of ML-DSA the statement could be held up, since according to our comparison it faster then conventional curves but considerably slower than EdDSA.

But with respect to space requirements, I don't think the statement can be held: For example for 128-bit security, Dilithium signatures are about 37x larger than EC signatures, while SLH-DSA small parameters only bring another factor of 3. 

If we compare ML-DSA with RSA (3072 bit), then ML-DSA signatures are 6x larger.

```suggestion
execution time requirements that is close to that of traditional ECC signature schemes.
Regarding the size of signatures and public keys, though, ML-DSA has far greater requirements than traditional schemes like EC-based or even RSA
```

draft-wussler-openpgp-pqc.md
```diff
@@ -1565,84 +1489,88 @@ setting, the adversary has to distinguish a set of challenge keys
...
+signature schemes. Implementers may want to offer SLH-DSA for applications
 where a higher degree of trust in the signature scheme is required. However,
-SPHINCS+ has performance characteristics in terms of execution time of the
+SLH-DSA has performance characteristics in terms of execution time of the
 signature generation as well as space requirements for the signature that can
 be, depending on the parameter choice, far greater than those of traditional or
```
falko-strenzke, line 1546:
```suggestion
even greater than those of
```

draft-wussler-openpgp-pqc.md
```diff
@@ -1565,84 +1489,88 @@ setting, the adversary has to distinguish a set of challenge keys
...
-Generally, it can be said that Dilithium + ECC provides a performance in terms
-of execution time and space requirements that is close to that of traditional
-ECC signature schemes. Implementers may want to offer SPHINCS+ for applications
+Generally, it can be said that ML-DSA + ECC provides a performance in terms of
+execution time and space requirements that is close to that of traditional ECC
+signature schemes. Implementers may want to offer SLH-DSA for applications
 where a higher degree of trust in the signature scheme is required. However,
-SPHINCS+ has performance characteristics in terms of execution time of the
+SLH-DSA has performance characteristics in terms of execution time of the
 signature generation as well as space requirements for the signature that can
```
falko-strenzke, line 1545:
```suggestion
signature generation as well as space requirements for the signature that are
```

falko-strenzke (approved):

Issue #69, opened 2023-11-08
Abbreviation MB appears out of nowhere, is not used again


teythoon:
Section 5.2.2 Key Combiner has:

> MB = KMAC256(domSeparation, encData, oBits, customizationString)

But MB is not motivated, nor used ever again.

falko-strenzke:
Will be fixed by #80.

falko-strenzke:
Fixed in #80

Issue #68, opened 2023-11-08
Consistency in the KEM descriptions


teythoon:
Section 5.1.2 about ML-KEM starts with

> 1. Extract the encapsulation key mlkemPublicKey that is part of the recipient's composite public key

but in the description of X*-KEM and ECDH-KEM, the recipients pubic key appears out of nowhere.  I think this should be made consistent, one way or the other.

falko-strenzke:
Now addressed in https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/80

falko-strenzke:
Closing as fixed in #80

Issue #15, opened 2023-01-09
Consistency with respect to wording "extraction" of component key


falko-strenzke:
# Falko's original remark

In Section 4.1.2 we state

> Extract the component public key kyberPublicKey that is part of the recipient's composite public key

This seems inconsistent in wording with the descriptions for the EC-KEMs, where no mentioning of the extraction of the component public key from the composite key is made. IMO it should be mentioned in for each scheme or for none.

# Aron's comment on this (from previous issue)

> In Section 4.1.2
>This seems inconsistent in wording with the descriptions for the EC-KEMs, 

In the ECC sections we mention 
```...where R is the component public key eccPublicKey...```
I would consider removing point (1) from 4.1.2 and rewrite
``` 
2. Invoke (kyberCipherText, keyShare) <- kyberKem.encap(kyberPublicKey) 
   where kyberPublicKey is extracted from recipient's composite public key
```

falko-strenzke:
Meeting 2023-01-19:

* Aron & Falko: just about consistency, will be aligned

falko-strenzke:
Meeting 2023-06-05:
* AW will work on this

falko-strenzke:
will be fixed by #80

falko-strenzke:
fixed via #80

PR #80, opened 2024-02-05
Clarifications mostly to KEM specification

falko-strenzke:
These are all mere clarifications. 

Note that X25519 is defined not with a scalar input, but with [an octet string as input](https://datatracker.ietf.org/doc/html/rfc7748#section-6.1).

"the fixed 4 byte value 0x00000001" did specify a number, not a sequence of bytes. The endianess of the encoding would still have to be given here.

The remaining changes should be obvious.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -303,15 +303,17 @@ based on these mathematical problems is endangered by the advent of quantum
 computers, there is a need to extend OpenPGP by algorithms that remain secure
 in the presence of quantum computers.
 
-Such cryptographic algorithms are referred to as post-quantum cryptography.
-The algorithms defined in this extension were chosen for standardization by the
+Such cryptographic algorithms are referred to as post-quantum cryptography.  The
+algorithms defined in this extension were chosen for standardization by the
 National Institute of Standards and Technology (NIST) in mid 2022
 {{NISTIR-8413}} as the result of the NIST Post-Quantum Cryptography
-Standardization process initiated in 2016 {{NIST-PQC}}. Namely, these are
-ML-KEM (formerly CRYSTALS-Kyber) as a Key Encapsulation Mechanism (KEM), a KEM
-being a modern building block for public-key encryption, and ML-DSA (formerly
-CRYSTALS-Dilithium) as well as SLH-DSA (formerly SPHINCS+) as signature
-schemes.
+Standardization process initiated in 2016 {{NIST-PQC}}. Namely, these are ML-KEM
+("Module Lattice Key Encapsulation Mechanism", submitted to the NIST process as
+"CRYSTALS-Kyber") as a Key Encapsulation Mechanism (KEM), a KEM being a modern
+building block for public-key encryption, and ML-DSA ("Module Lattice Digital
+Signature Algorithm", submitted to the NIST process as "CRYSTALS-Dilithium") as
+well as SLH-DSA ("Stateless Hash Digital Signature Algorithm", submitted to the NIST process as
+"SPHINCS+") as signature schemes.
 
 For the two ML-* schemes, this document follows the conservative strategy to
 deploy post-quantum in combination with traditional schemes such that the
@@ -744,8 +746,9 @@ The operation `x448Kem.Decaps()` is defined as follows:
 
 The operation `ecdhKem.Encaps()` is defined as follows:
 
- 1. Generate an ephemeral key pair {`v`, `V=vG`} as defined in
-    {{SP800-186}} or {{RFC5639}} where `v` is a random scalar
+ 1. Generate an ephemeral key pair {`v`, `V=vG`} as defined in {{SP800-186}} or
+    {{RFC5639}} where `v` is a random scalar with `0 < v < n`, `n` being the
+    base point order of the elliptic curve domain parameters
 
  2. Compute the shared point `S = vR`, where `R` is the component public key
     `eccPublicKey`, according to {{SP800-186}} or {{RFC5639}}
@@ -962,7 +965,7 @@ scheme is as follows:
  10. Output the algorithm specific part of the PKESK as
      `eccCipherText || mlkemCipherText (|| symAlgId) || len(C) || C`, where
      both `symAlgId` and `len(C)` are single octet fields and `symAlgId`
-     denotes the symmetric algorithm used and is present only for v3 PKESK
+     denotes the symmetric algorithm used and is present only for a v3 PKESK
 
 ### Decryption procedure
 
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -864,7 +864,8 @@ For the composite KEM schemes defined in {{kem-alg-specs}} the following
 procedure MUST be used to compute the KEK that wraps a session key. The
 construction is a one-step key derivation function compliant to {{SP800-56C}}
 Section 4, based on KMAC256 {{SP800-185}}. It is given by the following
-algorithm.
+algorithm, which computes the key encryption key `KEK` that is used to wrap,
+i.e., encrypt, the session key.
 
     //   multiKeyCombine(eccKeyShare, eccCipherText,
     //                   mlkemKeyShare, mlkemCipherText,
@@ -888,7 +889,8 @@ algorithm.
     mlkemData = mlkemKeyShare || mlkemCipherText
     encData = counter || eccData || mlkemData || fixedInfo
 
-    MB = KMAC256(domSeparation, encData, oBits, customizationString)
+    KEK = KMAC256(domSeparation, encData, oBits, customizationString)
+    return KEK
 
 Here, the parameters to KMAC256 appear in the order as specified in
 {{SP800-186}}, Section 4, i.e., the key `K`, main input data `X`, requested
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -697,7 +697,7 @@ the base point of Curve25519.
 The operation `x25519Kem.Encaps()` is defined as follows:
 
  1. Generate an ephemeral key pair {`v`, `V`} via `V = X25519(v,U(P))` where `v`
-    is a random scalar
+    is a randomly generated octet string with a length of 32 octets
 
  2. Compute the shared coordinate `X = X25519(v, R)` where `R` is the public key
     `eccPublicKey`
@@ -724,7 +724,7 @@ of Curve448.
 The operation `x448.Encaps()` is defined as follows:
 
  1. Generate an ephemeral key pair {`v`, `V`} via `V = X448(v,U(P))` where `v`
-    is a random scalar
+    is a randomly generated octet string with a length 56 octets
 
  2. Compute the shared coordinate `X = X448(v, R)` where `R` is the public key
     `eccPublicKey`
@@ -884,7 +884,7 @@ algorithm.
     //   Constants:
     //   domSeparation       - the UTF-8 encoding of the string
     //                         "OpenPGPCompositeKeyDerivationFunction"
-    //   counter             - the fixed 4 byte value 0x00000001
+    //   counter             - the 4 byte value 00 00 00 01
     //   customizationString - the UTF-8 encoding of the string "KDF"
 
     eccData = eccKeyShare || eccCipherText
@@ -892,6 +892,9 @@ algorithm.
     encData = counter || eccData || mlkemData || fixedInfo
 
     MB = KMAC256(domSeparation, encData, oBits, customizationString)
+
+Here, the inputs to KMAC256 have the meaning key, main input data, requested
+output length, and optional customization string in that order.
 
 Note that the values `eccKeyShare` defined in {{ecc-kem}} and `mlkemKeyShare`
 defined in {{mlkem-ops}} already use the relative ciphertext in the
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -303,15 +303,17 @@ based on these mathematical problems is endangered by the advent of quantum
 computers, there is a need to extend OpenPGP by algorithms that remain secure
 in the presence of quantum computers.
 
-Such cryptographic algorithms are referred to as post-quantum cryptography.
-The algorithms defined in this extension were chosen for standardization by the
+Such cryptographic algorithms are referred to as post-quantum cryptography.  The
+algorithms defined in this extension were chosen for standardization by the
```
dkg, line 307: this seems to just be a textual reflow.  is there a reason to include it in this change?

draft-ietf-openpgp-pqc.md
```diff
@@ -724,7 +726,7 @@ of Curve448.
 The operation `x448.Encaps()` is defined as follows:
 
  1. Generate an ephemeral key pair {`v`, `V`} via `V = X448(v,U(P))` where `v`
-    is a random scalar
+    is a randomly generated octet string with a length 56 octets
```
dkg, line 729: for X25519, it says "with a length *of* 32 octets".  This omits the "of".  maybe align them?

draft-ietf-openpgp-pqc.md
```diff
@@ -303,15 +303,17 @@ based on these mathematical problems is endangered by the advent of quantum
...
-being a modern building block for public-key encryption, and ML-DSA (formerly
-CRYSTALS-Dilithium) as well as SLH-DSA (formerly SPHINCS+) as signature
-schemes.
+Standardization process initiated in 2016 {{NIST-PQC}}. Namely, these are ML-KEM
+("Module Lattice Key Encapsulation Mechanism", submitted to the NIST process as
+"CRYSTALS-Kyber") as a Key Encapsulation Mechanism (KEM), a KEM being a modern
+building block for public-key encryption, and ML-DSA ("Module Lattice Digital
+Signature Algorithm", submitted to the NIST process as "CRYSTALS-Dilithium") as
+well as SLH-DSA ("Stateless Hash Digital Signature Algorithm", submitted to the NIST process as
+"SPHINCS+") as signature schemes.
```
dkg, line 316: It's not clear to me why "submitted to the NIST process as…" needs to be here for any of these three schemes, particularly in the introduction.  This is historical data at this point, right?  and the schemes as submitted don't necessarily match the schemes as standardized or used here.

If the goal is that someone searching an archive of IETF documents for "Dilithium", etc can find this draft, maybe just create a "historical notes" appendix or something.  otherwise, i think just dropping the "submitted to the NIST process as …" texts would also be fine.

draft-ietf-openpgp-pqc.md
```diff
@@ -965,7 +974,8 @@ scheme is as follows:
 
  1. Take the matching PKESK and own secret key packet as input
 
- 2. From the PKESK extract the algorithm ID and the `encryptedKey`
+ 2. From the PKESK extract the algorithm ID and the `encryptedKey`, i.e. the
```
dkg, line 977: nit-pick: comma after "i.e."

draft-ietf-openpgp-pqc.md
```diff
@@ -1016,10 +1026,12 @@ The algorithm-specific fields consists of:
    - The wrapped session key represented as an octet string, i.e., the output
      of the encryption procedure described in {{ecc-mlkem-encryption}}.
 
-Note that unlike most public-key algorithms, in the case of a v3 PKESK packet,
-the symmetric algorithm identifier is not encrypted.  Instead, it is prepended
-to the encrypted session key in plaintext.  In this case, the symmetric
-algorithm used MUST be AES-128, AES-192 or AES-256 (algorithm ID 7, 8 or 9).
+Note that unlike most public-key algorithms, for the ML-KEM composite schemes,
```
dkg, line 1029: Is "unlike most" really what we want to say here?  It's also true for the crypto-refresh's recommended X25519 and X448 public-key algorithms.  Maybe it's better to refer to the similiarity rather than to the difference?

draft-ietf-openpgp-pqc.md
```diff
@@ -1016,10 +1026,12 @@ The algorithm-specific fields consists of:
...
+in the case of a v3 PKESK packet, the symmetric algorithm identifier is not
+encrypted.  Instead, it is prepended to the encrypted session key in plaintext.
```
dkg, line 1031: is "prepended" the right way to say this?  it is prepended to `len(C)||C`, but it is after `eccCipherText||mlkemCipherText`.

draft-ietf-openpgp-pqc.md
```diff
@@ -1016,10 +1026,12 @@ The algorithm-specific fields consists of:
...
+In this case, the symmetric algorithm used MUST be AES-128, AES-192 or AES-256
+(algorithm ID 7, 8 or 9).
```
dkg, line 1033: Should we justify this?

I note that in crypto-refresh-13, the same concern is not explicitly justified in the text, but in commit 3eea8ee71215a2e1be24a3d45ac928aa098720a4 @twiss justifies it with:

> Since the symmetric algorithm ID is not encrypted nor authenticated in v3 PKESKs, mandate that AES is used, to prevent a (remote) risk of cross-algorithm attacks.

dkg (commented):
This looks like an overall improvement in the draft.  I've left a series of nit-picking comments that could improve it further though.  Thanks for working on it, @falko-strenzke !

draft-ietf-openpgp-pqc.md
```diff
@@ -893,6 +896,9 @@ algorithm.
 
     MB = KMAC256(domSeparation, encData, oBits, customizationString)
 
+Here, the inputs to KMAC256 have the meaning key, main input data, requested
```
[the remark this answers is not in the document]
falko-strenzke: > It should be made more clear that this is meant as a mapping between the KMAC terminology `KMAC256(K, X, L, S)` and the inputs that we pass in the combiner. I think explicitly naming the parameters `K, X, L, S` and referencing to Section 4.2 and 4.3 in SP800-185 will make it more clear.

OK, fixed.
> 
> Implementers will face the additional challenge of mapping these values to the MAC API of their chosen crypto library which can be unintuitive. For example, the nonce/iv will be the `customizationString`. It's probably out of scope for this document, but I'd not be surprised if this confuses some people. So my question is: Can or should we add some hints in the document?

I don't see that we can clarify the use of library interfaces for implementers on that level. Then we would have do that for each and every cryptographic operation that we mention. Each time this would involve lengthy text full of speculation how a library might have to be used. I think this would do more harm than good in terms of clarity.

draft-ietf-openpgp-pqc.md
falko-strenzke: Yes, a good suggestion, done.

draft-ietf-openpgp-pqc.md
falko-strenzke: I tried to describe it exactly now.

draft-ietf-openpgp-pqc.md
falko-strenzke: Fixed.

draft-ietf-openpgp-pqc.md
falko-strenzke: I agree, that this is information is superfluous in the introduction. I replaced it by the references to the FIPS standards for each scheme.

draft-ietf-openpgp-pqc.md
falko-strenzke: Yes, clearly the "of" was missing. Added it now.

draft-ietf-openpgp-pqc.md
falko-strenzke: It seems to have happened due to an inaccuracy of my editor (Vim) which made a reflow to limit the line to one character less than I specified. I reverted it.

draft-ietf-openpgp-pqc.md
```diff
@@ -956,7 +962,10 @@ scheme is as follows:
  9. Compute `C := AESKeyWrap(KEK, sessionKey)` with AES-256 as per {{RFC3394}}
     that includes a 64 bit integrity check
 
- 10. Output `eccCipherText || mlkemCipherText || len(C) || C`
+ 10. Output the algorithm specific part of the PKESK as
+     `eccCipherText || mlkemCipherText (|| symAlgId) || len(C) || C`, where
```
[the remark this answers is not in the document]
falko-strenzke: Indeed. I reorganized Section 5.3.1 even a bit more.

draft-ietf-openpgp-pqc.md
```diff
@@ -956,7 +962,10 @@ scheme is as follows:
...
+     both `symAlgId` and `len(C)` are single octet fields and `symAlgId`
+     denotes the symmetric algorithm used and is present only for a v3 PKESK
```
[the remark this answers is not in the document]
falko-strenzke: Fixed.

draft-ietf-openpgp-pqc.md
falko-strenzke: I would even go so far to say that the chosen approach is not 100% waterproof. It prevents cross-algorithm attacks, but not necessarily cross-key-size attacks in all cases. If an implementation accepts a 256 bit value for an AES-128 key by dropping the second half, then such an attack is possible. I would clearly say that a secure and robust implemention should not accept oversized keys, but still: with the chosen approach we increase the attack surface without any justifiable reason, except that we want align with the crypto-refresh.

wussler: In our implementations we already check the key size matching the algorithm, and error our otherwise. I see no issue in mandating this check.

draft-ietf-openpgp-pqc.md
```diff
@@ -956,7 +960,10 @@ scheme is as follows:
  9. Compute `C := AESKeyWrap(KEK, sessionKey)` with AES-256 as per {{RFC3394}}
     that includes a 64 bit integrity check
 
- 10. Output `eccCipherText || mlkemCipherText || len(C) || C`
+ 10. Output the algorithm specific part of the PKESK as
+     `eccCipherText || mlkemCipherText (|| symAlgId) || len(C) || C`, where
+     both `symAlgId` and `len(C)` are single octet fields and `symAlgId`
+     denotes the symmetric algorithm ID used and is present only for a v3 PKESK
```
wussler, line 966:
```suggestion
     denotes the symmetric algorithm ID used and is present only for a v3 PKESK.
```

draft-ietf-openpgp-pqc.md
```diff
@@ -998,28 +1007,29 @@ scheme is as follows:
...
 
-   - The wrapped session key represented as an octet string, i.e., the output
-     of the encryption procedure described in {{ecc-mlkem-encryption}}.
+Note that like in the case of the algorithms X25519 and X448 specified in
+{{I-D.ietf-openpgp-crypto-refresh}}, for the ML-KEM composite schemes, in the
+case of a v3 PKESK packet, the symmetric algorithm identifier is not encrypted.
+Instead, it is placed in plaintext after the `mlkemCipherText` and before the
+length octet preceding the wrapped session key.  In the case of v3 PKESK packets
+for ML-KEM composite schemes, the symmetric algorithm used MUST be AES-128,
+AES-192 or AES-256 (algorithm ID 7, 8 or 9).
```
wussler, line 1031:
```suggestion
AES-192 or AES-256 (algorithm ID 7, 8 or 9).

An implementation MUST check if the key length matches the key algorithm identifier, and abort otherwise.
```

This would fully prevent cross-algorithm attacks

draft-ietf-openpgp-pqc.md
falko-strenzke: There are no periods after the preceding sentences in the algorithm specification. If we insert it here, we should do so in all algorithms in all steps.

wussler: You are right, I had seen that somewhere else we have a period on the last one, but we should avoid here

draft-ietf-openpgp-pqc.md
falko-strenzke: Good suggestion, I adopted a slightly different variant of the sentence.

draft-ietf-openpgp-pqc.md
```diff
@@ -699,7 +699,7 @@ The operation `x25519Kem.Encaps()` is defined as follows:
     is a randomly generated octet string with a length of 32 octets
 
  2. Compute the shared coordinate `X = X25519(v, R)` where `R` is the public key
-    `eccPublicKey`
+    `eccPublicKey`, i.e. the recipient's public key
```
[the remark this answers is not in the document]
falko-strenzke: OK, fixed

draft-ietf-openpgp-pqc.md
```diff
@@ -726,7 +726,7 @@ The operation `x448.Encaps()` is defined as follows:
     is a randomly generated octet string with a length of 56 octets
 
  2. Compute the shared coordinate `X = X448(v, R)` where `R` is the public key
-    `eccPublicKey`
+    `eccPublicKey`, i.e. the recipient's public key
```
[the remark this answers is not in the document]
falko-strenzke: OK, fixed

fluppe2 (approved):

falko-strenzke:
fixes #69

falko-strenzke:
fixes #68

falko-strenzke:
fixes #15 

@wussler Could you also review at least the last commit renaming "MB = KMAC(..." as I made this change after @fluppe2 and @TJ-91 had approved?

wussler (approved):
The whole PR looks good to me. Thanks!

PR #81, opened 2024-02-05
remove the binding between v6 and PQC encryption keys in the …

TJ-91:
…Key Generation Strategies.

The text still referred to the state of the document when it was mandated to use v6 with PQC. This is now relaxed for encryption subkeys.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -1395,7 +1395,7 @@ It is REQUIRED to generate fresh secrets when generating PQ(/T) keys. Reusing
 key material from existing ECC keys in PQ(/T) keys does not provide backwards
 compatibility, and the fingerprint will differ.
 
-An OpenPGP (v6) certificate is composed of a certification-capable primary key
+An OpenPGP certificate is composed of a certification-capable primary key
 and one or more subkeys for signature, encryption, and authentication.
 Two migration strategies are recommended:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -1395,19 +1395,20 @@ It is REQUIRED to generate fresh secrets when generating PQ(/T) keys. Reusing
...
 understanding PQ(/T) certificates can use the legacy certificate, while
 PQ(/T)-capable implementations will prefer the newer certificate. This allows
-having an older v4 or v6 ECC certificate for compatibility and a v6 PQ(/T)
+having an older v4 or v6 certificate for compatibility and a v6 PQ(/T)
 certificate, at a greater complexity in key distribution.
 
-2. Attach PQ(/T) encryption and signature subkeys to an existing v6 ECC
-certificate. Implementations understanding PQ(/T) will be able to parse and use
+2. Attach PQ(/T) encryption subkeys to an existing traditional OpenPGP certificate.
+In the case of a v6 certificate, also PQ(/T) signature keys may be attached.
```
falko-strenzke, line 1410:
```suggestion
v6 PQ(/T) signature keys may also be attached to a certificate with a v4 or v6 primary key. The PQ(/T) signature keys may then be used whenever there is sufficient evidence that the recipient will be able to verify such signatures.
```

I understand that this statement may be controversial, because of the questionable security gain in PQ signature subkeys under a traditional primary key. But since anyone who receives such such a public key in an authenticated way and then stores it until the quantum computer era, can still be sure to have authentic PQ signature keys of that party.

falko-strenzke: Oh, yes, clearly you can't attach a v6 subkey to a v4 primary key. You are right, this is an entirely different topic I brought up here. Probably it will surface anyway in the upcoming discussions, so I don't think we necessarily have to create an issue now.

wussler (approved):
Looks good to me, also good hint of specifying v6 + sigs

falko-strenzke (approved):

Issue #83, opened 2024-02-07
Please avoid squatting on codepoints


dkg:
The draft currently stakes out claims for codepoints in the active range of asymmetric keys, though the values are still in flux.  That seems dangerous.

If we want concrete values for interoperability testing, feel free to use experimental codepoints (100-110) while the draft is still in flux.

We'll stabilize on formal codepoints later in the process once we know what we're getting into.

falko-strenzke:
I am not entirely sure how to understand this proposal, as the draft currently defines 6 KEM algos and 8 signature algos, so a total of 14 altogether which is more than fits into the range 100-110. To be closest to the current proposal would thus be to use 100 - 113.

The – possibly more consistent – alternative would be to start from 111 in order to spare the anyway insufficient range of experimental code points entirely.

dkg:
Please see #86 for one possible way to use experimental codepoints while the draft is still in flux

dkg:
@TJ-91 i like your suggestion to just leave the MAY algorithms explicitly `TBD` and only use experimental codepoints for the SHOULD and MUST algorithms.  Would you prepare a pull request with that?

PR #86, opened 2024-02-08
use experimental codepoints for now, during interop testing

dkg:
Closes: #83


dkg:
@TJ-91 suggests on list simply not assigning codepoints to the MAY algorithms, rather than having overlap.  This also seems fine to me.

wussler:
Closing in favor of #87

PR #87, opened 2024-02-12
assign experimental/private codepoints; leave out MAY algorithms

### LINKED ISSUE #83: Please avoid squatting on codepoints

TJ-91:
fixes #83. this is an alternative to PR #86.

I only assign IDs to "MUST" and "SHOULD" algorithms.
In cases where a referenced ID is not assigned, I write `TBD (Full Algorithm Name)` to keep the reference in tact.

In cases where there are multiple IDs listed and the IDs are only there for convenience and the meaning of the table is still clear, I only write "TBD" to keep the draft more readable. For example in `Table 12: ML-DSA parameters and artifact lengths in octets`.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -1143,8 +1143,8 @@ support the matching hash algorithm.
 {: title="Binding between ML-DSA and signature data digest" #tab-mldsa-hash}
 Algorithm ID reference | Hash function | Hash function ID reference
 ----------------------:| ------------- | --------------------------
-TBD                    | SHA3-256      | 12
-TBD                    | SHA3-512      | 14
+TBD (ML-DSA-65 IDs)    | SHA3-256      | 12
+TBD (ML-DSA-87 IDs)    | SHA3-512      | 14
 
 ### Key generation procedure {#ecc-mldsa-generation}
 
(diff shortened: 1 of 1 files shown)


falko-strenzke (changes requested):
Thanks for coming up with the solution @TJ-91 and @dkg and this PR. The only thing that I am a bit concerned about right now is that I anticipate that we will rather prioritise SLH-DSA-SHAKE over -SHA2, and thus the meaning of that code point will change again, potentially causing problems during interop testing if one implementation misses to adopt that change. So maybe we wait a bit with the assignment of any code points for SLH-DSA right now?

falko-strenzke:
> We still have codepoints left, so we can just use the next one for SHAKE if it comes to that. I'd be fine with both options.

Yes, I think we can worry about that later.

falko-strenzke (approved):

wussler:
May sound silly, but what about using 105-109 instead? Most experimental algos I know of are already using 100-102 and this could be a source of conflict

falko-strenzke:
> May sound silly, but what about using 105-109 instead? Most experimental algos I know of are already using 100-102 and this could be a source of conflict

Then that makes sense I think. Clearly, none of these solutions are robust – as soon as anyone else arrives at the same solution there will be a conflict...

falko-strenzke (approved):

wussler (approved):
Thank you!

dkg:
Thanks for this, @TJ-91 !  I think this is the right way forward.

PR #96, opened 2024-02-28
Add test vectors

wussler:
Add test vectors for 
- V6 Ed25519 key with X25519+ML-KEM-786 encryption subkey
- V4 Ed25519 key with X25519+ML-KEM-786 encryption subkey


wussler:
I ended up removing the checksum everywhere, they are test vectors anyway

draft-ietf-openpgp-pqc.md
```diff
@@ -1707,6 +1708,161 @@ Andreas Huelsing (TU Eindhoven)
...
+# Test Vectors
+
+To help implementing this specification a set of non-normative examples follow here.
+
+## Sample v6 PQC Subkey Artifacts
+
+Here is a Private Key consisting of:
+
+- A v6 Ed25519 Private-Key packet
+- A User ID packet
```
dkg, line 1720: v6 certificates need a direct key self-sig on the primary key (see [crypto-refresh §10.1.1](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-13.html#name-openpgp-v6-certificate-stru))

wussler: Right, missed this in the description. I will push now a commit to fix

dkg:
can i discourage you from using references to Golang in the user ID?  it makes the test vector seem implementation-specific, and the goal of a test vector is promoting cross-implementation interoperability.

dkg:
Let me review the purpose of the v4 EdDSALegacy+PQ/T test vector (`test-vectors/v4-eddsa-sample*`):

- Some implementations cannot parse v6 or crypto-refresh Ed25519, and we want those implementations to be able to encrypt to this key.
- An implementation that cannot parse v6 will surely also not be able to parse or encrypt to the ML-KEM-768+X25519 
- We offer the PQ/T encryption-capable subkey so that implementations which *can* encrypt to PQ/T can do it.

In summary, this test vector represents a "compromise key" that is supposed to interoperate with legacy implementations, while *also* being usable by fully up-to-date implementations in 2024 that encrypt with the best-known mechanisms.

If that's the goal, then i draw the following conclusions:

- the "compromise key" should advertise SEIPDv2.  Legacy implementations don't need to encrypt using SEIPDv2, but there is no reason to not advertise it for use by up-to-date implementations.  It should also include a "Preferred AEAD Ciphersuites" subpacket.
- the "compromise key" should also have an ECDH encryption-capable subkey, presumably using Curve25519Legacy, so that legacy implementations can encrypt to it.  I recommend having this second encryption-capable subkey with a creation timestamp and subkey binding signature with earlier timestamps than the PQ/T encryption-capable subkey.

Test vectors have a way of setting defacto standards, and it would be great to start off with something as close to what we think implementations should be generating as possible.

Having the "compromise key" test vector be as plausible as possible will also help us to evaluate whether it actually serves the "compromise" purpose or not, by testing using it with legacy implementations (similar to [the interop test suite's "Mock PQ Subkey" test](https://tests.sequoia-pgp.org/#Mock_PQ_subkey)).

It's possible that my two requests above aren't the only things needed to make a good "compromise key".   Can you update this test vector to be as close as possible to what you think would be a plausible "compromise key"?

dkg:
@TJ-91 wrote:
> There is the possibility to leave out v6 keys, v6 PKESK and v2 SEIPD entirely (parsing and generating), but to include PQ/T encryption.

Sure, but i think the goal should be interoperating with actual legacy software, not interoperating with an implementer practicing deliberate avoidance.  There is no way to interoperate with an implementer who is trying to not interoperate, so there is no point in targeting them.  And the v6 key and sig formats, and v2 SEIPD are now stable, well-specified, and simple compared to implementing anything like ML-KEM.

wussler:
Okay, I'll do the following changes to the vectors:
- Change the UserIDs
- Add the SEIPD flags to the v4 key 
- Add both a SEIPDv1 and a SEIPDv2 encrypted message to the v4 key
- Add an ECDH subkey to the v4 key
- Add the CRC24 on the v4 armored key

Regarding

> I recommend having this second encryption-capable subkey with a creation timestamp and subkey binding signature with earlier timestamps than the PQ/T encryption-capable subkey.

Note that in the spec we explicitly state

> Implementations SHOULD prefer PQ(/T) keys when multiple options are available.

Not setting an earlier date can also be a test for this case.

falko-strenzke (approved):

PR #99, opened 2024-03-21
update SLH-DSA to IETF-119 proposal

falko-strenzke:
For technical reasons, namely being able to proceed with other changes affecting larger portions of the draft without creating any conflicts, the editors would like to merge this PR as soon as it is found to make the correct transformation to the new three SLH-DSA-SHAKE code points (removing the parametrization) in a  *technical sense* (which has an impact on the document structure and the textual reference to the algorithm) .

The criterion for merging it should not be that there is already consensus on the exact number of SLH-DSA code points or the text for their justification in the "Additional Considerations" section. That can be agreed on in the further discussion on the list and then we can open another PR for the required changes.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -1583,24 +1583,37 @@ Regarding the size of signatures and public keys, though, ML-DSA has far greater
 requirements than traditional schemes like EC-based or even RSA
 signature schemes.
 
-Implementers may want to offer SLH-DSA-SHAKE for applications where the milder
-security assumptions of a hash-based signature scheme are required – and thus a
+Implementers may want to offer SLH-DSA-SHAKE for applications where the weaker
+security assumptions of a hash-based signature scheme are required – namely only the 2nd preimage resistance of a hash function – and thus a
 potentially higher degree of trust in the long-term security of signatures is
 achieved. However, SLH-DSA-SHAKE has performance characteristics in terms of
 execution time of the signature generation as well as space requirements for the
 signature that are even greater than those of ML-DSA + ECC signature schemes.
 
 Pertaining to the execution time, the particularly costly operation in
-SLH-DSA-SHAKE is the signature generation. In order to achieve short signature
+SLH-DSA-SHAKE is the signature generation. Depending on the parameter
+set, it can range from approximately the one hundred fold to more than the two
+thousand fold of that of ML-DSA-87. 
+These number are based on the performance measurements published in the NIST submissions for SLH-DSA-SHAKE and ML-DSA.
+In order to achieve fast signature
 generation times, the algorithm SLH-DSA-SHAKE-128f ("f" standing for "fast")
 should be chosen. This comes at the expense of a larger signature size.  This
-choice can be relevant in applications where mass signing occurs.
+choice can be relevant in applications where mass signing occurs or a small
+latency is required.
 
 In order to minimize the space requirements of an SLH-DSA-SHAKE signature, an
 algorithm ID with the name ending in "s" for "small" should be chosen. This comes
 at the expense of a longer signature generation time. In particular,
-SLH-DSA-SHAKE-128s achieves the smallest possible signature size. Where a higher
+SLH-DSA-SHAKE-128s achieves the smallest possible signature size, which is about
+the double size of an ML-DSA-87 signature. Where a higher
 security level than 128 bit is needed, SLH-DSA-SHAKE-256s can be used.
+
+Unlike the signature generation time, the signature verification time of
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -1632,61 +1571,52 @@ Implementations may also make more nuanced decisions.
...
-- Registry name: `OpenPGP SLH-DSA-SHAKE parameters`
+signature schemes.
 
-  Registration procedure: SPECIFICATION REQUIRED [RFC8126]
+Implementers may want to offer SLH-DSA-SHAKE for applications where the milder
+security assumptions of a hash-based signature scheme are required – and thus a
+potentially higher degree of trust in the long-term security of signatures is
+achieved. However, SLH-DSA-SHAKE has performance characteristics in terms of
+execution time of the signature generation as well as space requirements for the
+signature that are even greater than those of ML-DSA + ECC signature schemes.
```
dkg, line 1591: This is also a bit of a messy diff to read because it seems to be mainly a re-flow of the text, with only a few minor adjustments.   Having cleaner semantic diffs is one of the reasons that the crypto-refresh markdown source was formatted with one sentence per line.  That said, modulo the comments above, i think the underlying changes are generally OK here.

draft-ietf-openpgp-pqc.md
```diff
@@ -1632,61 +1571,52 @@ Implementations may also make more nuanced decisions.
...
-
-  Registration procedure: SPECIFICATION REQUIRED [RFC8126]
-
-  The registry contains the values defined in {{slhdsa-param-sha2}} in this document.
-
-- Registry name: `OpenPGP SLH-DSA-SHAKE parameters`
+signature schemes.
 
-  Registration procedure: SPECIFICATION REQUIRED [RFC8126]
+Implementers may want to offer SLH-DSA-SHAKE for applications where the milder
```
dkg, line 1586: It's not clear to me exactly what "milder" security assumptions are supposed to mean.  For example, If it means "less stringent threat model" , then that suggests that these signatures are less robust.  But if it means "less scary threat model", then that suggests that these signatures are more robust.

dkg (approved):
This seems like a reasonable next step to me.  I've noted some formatting concerns about the markdown that made the diff a bit harder to review than i would have liked, and some notes about language/vocab that i think can be cleaned up, and a few more details that i think could be added to the discussion, but none of those should block this step.  I'm happy to re-raise those issues after this is merged.

draft-ietf-openpgp-pqc.md
falko-strenzke: On sentence per line sounds more reasonable for the purpose of traceability of changes,  indeed. We'll discuss internally whether it could be an option to switch to that formatting. This would require to merge all bigger open PRs first, though.

draft-ietf-openpgp-pqc.md
falko-strenzke: Right, "milder" is not clear enough. I replaced it by "weaker". If it should not be clear how to understand this term, it is explained [here](https://crypto.stackexchange.com/questions/60059/comparing-stronger-vs-weaker-assumptions).

draft-ietf-openpgp-pqc.md
```diff
@@ -1632,61 +1571,52 @@ Implementations may also make more nuanced decisions.
...
+security assumptions of a hash-based signature scheme are required – and thus a
+potentially higher degree of trust in the long-term security of signatures is
+achieved. However, SLH-DSA-SHAKE has performance characteristics in terms of
+execution time of the signature generation as well as space requirements for the
+signature that are even greater than those of ML-DSA + ECC signature schemes.
+
+Pertaining to the execution time, the particularly costly operation in
+SLH-DSA-SHAKE is the signature generation. In order to achieve short signature
```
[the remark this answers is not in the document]
falko-strenzke: OK, I am adding information about the verification times, too.

draft-ietf-openpgp-pqc.md
```diff
@@ -1632,61 +1571,52 @@ Implementations may also make more nuanced decisions.
...
+TBD    | SLH-DSA-SHAKE-256s  | 64 octets public key ({{slhdsa-artifact-lengths}})                                                                     | 128 octets secret key ({{slhdsa-artifact-lengths}})                                                                    | 29792 octets signature ({{slhdsa-artifact-lengths}})                                                             | N/A                                                                                                                                                                                                    | {{slhdsa}}
```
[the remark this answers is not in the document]
falko-strenzke: Yes, I will respect that in the future :-) :+1:

PR #103, opened 2024-03-25
Adding missing changelog entry for slhdsa update

falko-strenzke:
Based on the branch of the PR https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/102 (but should not conflict with main)


falko-strenzke:
Makes more sense to base it on main, will create a new one.

PR #107, opened 2024-04-30
Remove NIST and Brainpool curve hybrids

wussler:
- Remove algorithms with NIST or Brainpool curves
- Remove ECDSA details
- Rename ECC to EdDSA or ECDH based on context
- Remove unneeded references
- Drop SEC1 point encoding

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -50,13 +50,9 @@ normative:
 
   RFC3394:
 
-  RFC8126:
-
   I-D.ietf-openpgp-crypto-refresh:
 
 informative:
-
-  RFC5639:
 
   NIST-PQC:
     target: https://csrc.nist.gov/projects/post-quantum-cryptography/post-quantum-cryptography-standardization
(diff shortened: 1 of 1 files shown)


wussler:
> Here for instance, the renaming "ECC" to "ECDH" or "EdDSA", resp., wasn't really necessary.

True, but we drop the ECDH-KEM (because it was the generic KEM for NIST/Brainpool curves), so we couldn't rename it without the split as it would have overloaded the term.

For ECDSA/EdDSA is what Johannes pointed out, just wanted to be more precise since now we implement only EdDSA.

> One really minor thing I would change in the Changelog: I wouldn't write "Moved NIST and Brainpool curve hybrids to a separate draft" since we have not yet done it (or have we?)

I can change it to "Removed NIST and Brainpool curve hybrids" if you think it's a good idea @TJ-91, since from the perspective of this draft this is what's happening.

draft-ietf-openpgp-pqc.md
```diff
@@ -1334,15 +1167,16 @@ TBD    | SLH-DSA-SHAKE-256s  | 64 octets public key ({{slhdsa-artifact-lengths}}
...
+- correction and completion of IANA instructions.
 
 ## draft-ietf-openpgp-pqc-02
-- Removed git rebase artifact
+- Removed git rebase artifact.
 
 ## draft-ietf-openpgp-pqc-03
-- Updated SLH-DSA by removing parametrization and restricting to three SLH-DSA-SHAKE algorithm code points
+- Updated SLH-DSA by removing parametrization and restricting to three SLH-DSA-SHAKE algorithm code points.
+- Moved NIST and Brainpool curve hybrids to a separate draft, dropped ECDSA from the current specification.
```
wussler, line 1179:
```suggestion
- Removed NIST and Brainpool curve hybrids, dropped ECDSA from the current specification.
```

PR #109, opened 2024-05-07
Assign SLH-DSA Codepoints

### LINKED ISSUE #105: Split off NIST/Brainpool composite draft
When the split is made, the following updates are necessary as well:

- remove the sentence "Algorithms indicated as MAY are not assigned a codepoint in the current state of the draft since there are not enough private/experimental code points available to cover all newly introduced public-key algorithm identifiers."

TJ-91:
I assigned code points 102 - 104 to the three SLH-DSA algorithms now. We can also do 104, 109, 110 or something like that, if there are concerns with using the lower experimental codepoints.

Fixes #105 by removing the now outdated sentence about unassigned code points.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -393,15 +393,13 @@ ID                    | Algorithm                          | Requirement | Defin
 ---------------------:| ---------------------------------- | ----------- | --------------------
 TBD (107 for testing) | ML-DSA-65 + Ed25519                | MUST        | {{ecc-mldsa}}
 TBD (108 for testing) | ML-DSA-87 + Ed448                  | SHOULD      | {{ecc-mldsa}}
-TBD (109 for testing) | SLH-DSA-SHAKE-128s                 | MAY         | {{slhdsa}}
-TBD                   | SLH-DSA-SHAKE-128f                 | MAY         | {{slhdsa}}
-TBD                   | SLH-DSA-SHAKE-256s                 | MAY         | {{slhdsa}}
+TBD (102 for testing) | SLH-DSA-SHAKE-128s                 | MAY         | {{slhdsa}}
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -863,9 +861,9 @@ This group of algorithms is henceforth referred to as "SLH-DSA-SHAKE code points
 {: title="SLH-DSA-SHAKE algorithm code points and the corresponding artifact lengths in octets." #slhdsa-artifact-lengths}
 Algorithm ID reference   |  SLH-DSA-SHAKE public key | SLH-DSA-SHAKE secret key | SLH-DSA-SHAKE signature
 ----------------------:  |  ------------------ | ------------------ | ------------------
-TBD (SLH-DSA-SHAKE-128s) |  32                 | 64                 | 7856
```
falko-strenzke, line 866: The column says "Algorithm ID reference", not "Algorithm ID". So unless the column label is changed, I would leave the textual algorithm label there.  The ID itself could maybe be added in a new column (leftmost)?

The same applies to the changes to the next table about the hash binding.

falko-strenzke: It's more reliable for now to keep the textual algorithm description, I think. The temporary inofficial IDs are not a reliable identification token. They might change and then readers might be confused when they see the next version.

falko-strenzke: I can also live with the proposed change, but I think for now being more explicit in those tables might be better.

falko-strenzke (approved):

wussler:
Do we need this until we can even produce a SLH-DSA compliant signature?

wussler:
I would avoid doing them now to leave some space for the other drafts in the adoption call. We can always see how it goes once we do have the SLH-DSA algorithms to make the test vectors.

Issue #91, opened 2024-02-13
binding signature hash for PQC encryption v4 keys


falko-strenzke:
Kai Engert kindly pointed out that Werner Koch had mentioned (on the LibrePGP list?) that with v4 PQC keys SHA-1 binding signatures would still be allowed. I currently don't see a problem with this, as I don't see how collission attacks against key binding signatures could be possible. The attacker would have to be able to supply one public key for which he has a collision and have that be signed by the victim.

Still, we might want to mandate that at least SHA-256 must be used in the binding signatures. Any opinions @wussler @fluppe2 @TJ-91 ?

kaie:
Werner had mentioned it in this message:
https://lists.gnupg.org/pipermail/librepgp-discuss/2024/000045.html

kaie:
Couldn't your draft include a statement like this?
"Applications that support PQC encryption keys defined in this document MUST ensure that the binding signature for the PQC subkey uses a SHA-256 signature or a stronger has. PQC encryption subkeys that use SHA-1 for the binding keys MUST be ignored."

dkg:
On Tue 2024-02-13 05:22:20 -0800, Falko Strenzke wrote:
> I don't see how collission attacks against key binding signatures
> could be possible. The attacker would have to be able to supply one
> public key for which he has a collision and have that be signed by the
> victim.

What about a scenario where an attacker offers the user a
decryption-capable subkey (e.g. by providing them with a smartcard)?

Even if we don't think this could ever be a plausible scenario, there is
a higher-level, simpler story that is useful to be able to tell, and it
goes like this:

- SHA-1 is never used in modern OpenPGP

Carving out random corner exceptions for one nuanced case or another
just weakens that story and makes it more challenging to offer clarity
for implementers and analysts.

dkg:
On Thu 2024-02-15 02:32:16 -0800, Johannes Roth wrote:
>> SHA-1 is never used in modern OpenPGP
>
> I think that is too ambiguous.

That's exactly my point.  It's already too ambiguous.  please don't make
it more ambiguous.

> Is v4 + PQC "modern"?

If by "v4 + PQC" you mean "following only the guidance in RFC 4880, but
with a post-quantum algorithm", this combination on its face is not
modern.  v4 was defined in RFC 4880 in 2007, over 16 years ago.  Even at
that time, it was known (via Wang et al.)  that SHA-1 was not everything
it was hoped it would be.

[...]

PR #110, opened 2024-05-14
symmetric algorithm preferences: clarify what happens in the absence …

### LINKED ISSUE #106: State that implicit preferred algorithm lists exist even when the corresponding subpacket is missing
Regarding the "Preferred AEAD Ciphersuites" subpacket and "Preferred Symmetric Ciphers for v1 SEIPD" subpacket, we need a clarification in Sec. "8.1. Symmetric Algorithms for SEIPD Packets". The crypto-refresh does not clearly state that the implicit lists of preferred algorithms are in force even when the respective subpacket is missing. We should clarify this at least in the PQC specification.

TJ-91:
…of the subpackets

Closes #106 

Not sure if the last sentence is really required, but the Crypto Refresh leaves a lot of room for implementations to determine v2 SEIPD support so it can't hurt.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -888,6 +888,8 @@ If the pair `AES-256` with `OCB` is not explicitly in the list of the "Preferred
 This is justified since `AES-256` and `OCB` are mandatory to implement.
 If the pair `AES-128` with `OCB` is also implicitly added to the list, it is added after the pair `AES-256` with `OCB`.
 
+When encrypting to an OpenPGP certificate that contains a PQ(/T) key, an implementation MUST NOT infer that the certificate's owner supports v2 SEIPD solely on the basis of the implicitly added AEAD algorithm pairs.
+
 ## Hash Algorithms for Key Binding Signatures
 
 Subkey binding signatures over algorithms described in this document and primary key binding signatures made by algorithms described in this document MUST NOT be made with `MD5`, `SHA-1`, or `RIPEMD-160`.


draft-ietf-openpgp-pqc.md
```diff
@@ -936,6 +936,12 @@ If the pair `AES-256` with `OCB` is not explicitly in the list of the "Preferred
 This is justified since `AES-256` and `OCB` are mandatory to implement.
 If the pair `AES-128` with `OCB` is also implicitly added to the list, it is added after the pair `AES-256` with `OCB`.
 
+When encrypting to a certificate that contains a PQ(/T) key
+and if one of the subpackets "Preferred Symmetric Ciphers for v1 SEIPD" or "Preferred AEAD Ciphersuites" is missing,
+an implementation SHOULD treat this as if the subpacket with an empty list exists
+and implicitly add algorithms as describhed in this section.
```
dkg, line 942: s/describhed/described/

draft-ietf-openpgp-pqc.md
```diff
@@ -936,6 +936,12 @@ If the pair `AES-256` with `OCB` is not explicitly in the list of the "Preferred
...
+Support for the v2 SEIPD feature MUST NOT be assumed on the basis of implicitly added AEAD algorithm pairs alone.
```
dkg, line 943: this final line is passive voice: who is doing the assuming or not assuming?

How about being clearer, like: 

> When encrypting to a OpenPGP certificate that contains a PQ(/T) key, an implementation MUST NOT infer that the certificate's owner supports v2 SEIPD solely on the basis of the implicitly added AEAD algorithm pairs.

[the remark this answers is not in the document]
dkg: i tend to agree with @wussler here that we probably don't need to say "lack of subpacket is the same as empty subpacket".  We also don't want to imply that any implementation should consider the presence of either encryption-support subpacket as the same as the appropriate bit in the Features subpacket.

dkg (changes requested):
I think this text can be further simplified while still getting across the main point.

I don't think implementers need reminding that "missing subpacket is the same as empty subpacket" -- in any case, if they do this might not be the right place to put it.

We're talking here about adding additional inference about supported encryption algorithms.  rather than stating what the implementation should *not* do (infer v2 SEIPD support on the basis of the PQ(/T) key alone), we should be explicit about what they *should* do: 

- look at the Features subpacket (today), or
- infer support for v2 SEIPD based on some survey of widely available implementations that support PQC (in the future, when such a survey is available).

wussler:
I am also happy with @dkg's proposed text, and I think it's now clear what to do in the case of missing subpackets. I also think here would be the right place to add a "forbid SED tag 9" for issue #91

dkg:
@TJ-91 i see what you're saying about the specifics of what is missing, but (a) the text you're proposing is a _lot_ of words to say "for these purposes, treat a missing subpacket the same as an empty subpacket", and (b) i don't really know what the alternative interpretation would be likely to be.  are you suggesting that an implementer might really think "if they have an empty AEAD ciphersuites subpacket, then i know i can use AES256-OCB, but if they have a missing AEAD ciphersuites subpacket, then i know i have to fall back to AES128-OCB"?  That would be very weird.

Maybe this would be best illustrated in a test vector, with a minimal certificate with a PQ/T encryption key, and a features packet that supports v2 SEIPD, and no AEAD Ciphersuites subpacket.  then the test vector could include an encrypted message that uses AES256-OCB .

draft-ietf-openpgp-pqc.md
falko-strenzke: The bottom line seems to be that there is consenus to only add the sentence that @dkg propsed above. In that case I ask @TJ-91 to update the PR accordingly. I am ready to close issue #106, which was the reason for the initially proposed changes in this PR, that was raised by me as irrelevant.

TJ-91: @wussler @falko-strenzke I rebased now with only that change

draft-ietf-openpgp-pqc.md
```diff
@@ -888,6 +888,8 @@ If the pair `AES-256` with `OCB` is not explicitly in the list of the "Preferred
 This is justified since `AES-256` and `OCB` are mandatory to implement.
 If the pair `AES-128` with `OCB` is also implicitly added to the list, it is added after the pair `AES-256` with `OCB`.
 
+When encrypting to an OpenPGP certificate that contains a PQ(/T) key, an implementation MUST NOT infer that the certificate's owner supports v2 SEIPD solely on the basis of the implicitly added AEAD algorithm pairs.
```
falko-strenzke, line 891: Since now we are going to remove the possibility of using ML-KEM and any other PQC encryption in v4 keys, isn't the whole topic obsolete now? 

PQC encryption implies a v6 key. A v6 key implies support for v6 PKESK and v2 SEIPD.

TJ-91: Technically the decision is not final for another month but yes, that's true. I didn't think of that.

I'll close this PR.

TJ-91:
We are planning to bind PQ(/T) encryption keys to v6 keys, making this obsolete.

PR #111, opened 2024-05-14
update pquip terminology reference

### LINKED ISSUE #58: Add/Update References to IETF pquip Documents
See https://datatracker.ietf.org/wg/pquip/documents/

* The terminology draft for PQ/T has been moved there (adopted by pquip WG), we need to reference the new document.
* https://datatracker.ietf.org/doc/draft-ar-pquip-pqc-engineers/ might also be a reference to consider

TJ-91:
Resolves #58

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -217,14 +217,17 @@ informative:
         - org: National Institute of Standards and Technology
       date: August 2023
 
-  draft-driscoll-pqt-hybrid-terminology:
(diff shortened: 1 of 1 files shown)


dkg:
This seems fine to me, but i would recommend relying on kramdown-rfc to generate the reference to the internet-draft directly, so that you don't have to maintain it.  I've offered #112 as an alternative.

I don't think you need to explicitly tag the version of the I-D you're referencing unless you expect that draft to actually radically change the terminology in use between versions.

Issue #67, opened 2023-11-08
Concerns regarding section 4.2, restrictions on T and PQ/T encryption to the same recipient


teythoon:
Curerntly, the spec states:

> [4.2. ](https://www.ietf.org/archive/id/draft-wussler-openpgp-pqc-03.html#section-4.2)[Parallel Public-Key Encryption](https://www.ietf.org/archive/id/draft-wussler-openpgp-pqc-03.html#name-parallel-public-key-encrypt)
>
> As explained in [Section 1.4.2](https://www.ietf.org/archive/id/draft-wussler-openpgp-pqc-03.html#non-composite-multi-alg), the OpenPGP protocol inherently supports parallel encryption to different keys of the same recipient. Implementations MUST NOT encrypt a message with a purely traditional public-key encryption key of a recipient if it is encrypted with a PQ/T key of the same recipient.

It is not quite clear to me what that means, and how implementations should enforce that restriction.  To be clear, I get the intention and I think it is a valid concern.

- Does that mean that if there is an OpenPGP certificate that has both a T and a PQ/T encryption subkey, an implementation MUST NOT encrypt to both subkeys?
- Does that mean that if an implementation has two OpenPGP certificates with the same user ID (or aliasing user ID, whatever that exactly means), where one has a T encryption subkey, and the other one has a PQ/T encryption subkey, an implementation MUST NOT encrypt to both subkeys?
- Does that mean that if an implementation has two OpenPGP certificates with different, non-aliasing user IDs, but the implementation or application knows somehow that the certificates belong to the same recipient, where one has a T encryption subkey, and the other one has a PQ/T encryption subkey, an implementation MUST NOT encrypt to both subkeys?

wussler:
We tried to tackle this in https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/20, but so far without success. We'll bring this to the list once we're clear on the larger topics

wussler:
Also duplicate of https://github.com/openpgp-pqc/draft-openpgp-pqc/issues/2

wussler:
Addressed in #120

PR #120, opened 2024-05-28
Rewrite parallel encryption guidance

wussler:
This PR addresses #2 and #67, removing the normative section about recipients, and making it an example for applications supporting it.
A review of all the terms "recipient" and "sender" was done, to remove the implicit reference to email

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -310,7 +310,7 @@ This specification introduces new cryptographic schemes, which can be categorize
 
  - PQ digital signature, namely SLH-DSA-SHAKE as a standalone cryptographic algorithm.
 
-For each of the composite schemes, this specification mandates that the communication party has to successfully perform the cryptographic algorithms for each of the component schemes used in a cryptographic message, in order for the message to be deciphered and considered as valid.
+For each of the composite schemes, this specification mandates that the consuming party has to successfully perform the cryptographic algorithms for each of the component schemes used in a cryptographic message, in order for the message to be deciphered and considered as valid.
 This means that all component signatures must be verified successfully in order to achieve a successful verification of the composite signature.
 In the case of the composite public-key decryption, each of the component KEM decapsulation operations must succeed.
 
@@ -322,7 +322,7 @@ These latter two cases realize non-composite combinations of signatures.
 {{multiple-signatures}} specifies how implementations should handle the verification of such combinations of signatures.
 
 Furthermore, the OpenPGP protocol also allows for parallel encryption to different keys.
-Accordingly, when a message is encrypted using multiple PKESK packages for different encryption keys, a non-composite, potentially multi-algorithm, public-key encryption is realized where the communication party has to decrypt only one of the PKESK packages in order to decrypt the message.
+Accordingly, when a message is encrypted using multiple PKESK packets for different encryption keys, a non-composite, potentially multi-algorithm, public-key encryption is realized where the consuming party has to decrypt only one of the PKESK packets in order to decrypt the message.
 
 # Supported Public Key Algorithms
 
@@ -531,7 +531,7 @@ The ML-KEM + ECDH composite public-key encryption schemes are built according to
 
  - The session key for content encryption is then wrapped as described in {{RFC3394}} using AES-256 as algorithm and the KEK as key.
 
- - The PKESK package's algorithm-specific parts are made up of the ML-KEM ciphertext, the ECDH ciphertext, and the wrapped session key.
+ - The PKESK packet's algorithm-specific parts are made up of the ML-KEM ciphertext, the ECDH ciphertext, and the wrapped session key.
 
 ### Fixed information {#kem-fixed-info}
 
@@ -868,7 +868,7 @@ The algorithm-specific part of the secret key consists of:
 ## Symmetric Algorithms for SEIPD Packets
 
 Implementations MUST implement `AES-256`.
-An implementation SHOULD use `AES-256` in the case of a v1 SEIPD packet, or `AES-256` with any available AEAD mode in the case of a v2 SEIPD packet, if all recipients indicate support for it (explicitly or implicitly).
+An implementation SHOULD use `AES-256` in the case of a v1 SEIPD packet, or `AES-256` with any available AEAD mode in the case of a v2 SEIPD packet, if all recipient certificates indicate support for it (explicitly or implicitly).
 
 A v4 or v6 certificate that contains a PQ(/T) key SHOULD include `AES-256` in the "Preferred Symmetric Ciphers for v1 SEIPD" subpacket.
 A v6 certificate that contains a PQ(/T) key SHOULD include the pair `AES-256` with `OCB` in the "Preferred AEAD Ciphersuites" subpacket.
@@ -995,7 +995,7 @@ This specification mandates support for `AES-256` for two reasons.
 First, `AES-KeyWrap` with `AES-256` is already part of the composite KEM construction.
 Second, some of the PQ(/T) algorithms target the security level of `AES-256`.
 
-For the same reasons, this specification further recommends the use of `AES-256` if it is supported by all communication parties, regardless of what the implementation would otherwise choose based on the communication parties' preferences.
+For the same reasons, this specification further recommends the use of `AES-256` if it is supported by all recipient certificates, regardless of what the implementation would otherwise choose based on the recipients' preferences.
 This recommendation should be understood as a clear and simple rule for the selection of `AES-256` for encryption.
 Implementations may also make more nuanced decisions.
 
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -322,7 +322,7 @@ These latter two cases realize non-composite combinations of signatures.
 {{multiple-signatures}} specifies how implementations should handle the verification of such combinations of signatures.
 
 Furthermore, the OpenPGP protocol also allows for parallel encryption to different keys.
-Accordingly, when a message is encrypted using multiple PKESK packages for different encryption keys, a non-composite multi-algorithm public-key encryption is realized where the communication party has to decrypt only one of the PKESK packages in order to decrypt the message.
+Accordingly, when a message is encrypted using multiple PKESK packages for different encryption keys, a non-composite, potentially multi-algorithm, public-key encryption is realized where the communication party has to decrypt only one of the PKESK packages in order to decrypt the message.
 
 # Supported Public Key Algorithms
 
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -321,8 +321,7 @@ Furthermore, multiple OpenPGP signatures may be combined on the application laye
 These latter two cases realize non-composite combinations of signatures.
 {{multiple-signatures}} specifies how implementations should handle the verification of such combinations of signatures.
 
-Furthermore, the OpenPGP protocol also allows for parallel encryption to different keys.
-Accordingly, when a message is encrypted using multiple PKESK packets for different encryption keys, a non-composite, potentially multi-algorithm, public-key encryption is realized where the consuming party has to decrypt only one of the PKESK packets in order to decrypt the message.
+Furthermore, the OpenPGP protocol also allows parallel encryption to different keys by using multiple PKESK packets, thus realizing non-composite multi-algorithm public-key encryption.
 
 # Supported Public Key Algorithms
 
@@ -369,7 +368,7 @@ This is achieved via KEM combination, i.e. both key encapsulations/decapsulation
 
 As explained in {{non-composite-multi-alg}}, the OpenPGP protocol inherently supports parallel encryption to different keys.
 Note that the confidentiality of a message is not post-quantum secure when encrypting to different keys if at least one key does not support PQ/T encryption schemes.
-In section {{pq-key-preference}} is explained how to deal with multiple key scenarios.
+In section {{pq-key-preference}} it is explained how to deal with multiple key scenarios.
 
 ## Composite Signatures
 
@@ -927,7 +926,7 @@ Two migration strategies are recommended:
 ## Security Aspects of Composite Signatures
 
 When multiple signatures are applied to a message, the question of the protocol's resistance against signature stripping attacks naturally arises.
-In a signature stripping attack, an adversary removes one or more of the signatures such that only a subset of the signatures remain in the message when verified.
+In a signature stripping attack, an adversary removes one or more of the signatures such that only a subset of the signatures remain in the message at the point when it is verified.
 This amounts to a downgrade attack that potentially reduces the value of the signature.
 It should be noted that the composite signature schemes specified in this draft are not subject to a signature stripping vulnerability.
 This is due to the fact that in any OpenPGP signature, the hashed meta data includes the signature algorithm ID, as specified in {{I-D.ietf-openpgp-crypto-refresh}}, Section 5.2.4.
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -894,13 +894,12 @@ Therefore various migration considerations must be taken into account, in partic
 ## Key preference {#pq-key-preference}
 
 Implementations SHOULD prefer PQ(/T) keys when multiple options are available.
-In case that a certificate has both a valid PQ/T and a valid traditional subkey, an implementation SHOULD use the PQ/T subkey only.
+When encrypting to a certificate that has both a valid PQ/T and a valid traditional encryption subkey, an implementation SHOULD use the PQ/T subkey only.
 Furthermore, if an application has any means to determine that encrypting to a PQ/T certificate and a traditional certificate is redundant, it should omit encrypting to the traditional certificate.
 
 As specified in {{composite-kem}}, the confidentiality of a message is not post-quantum secure when using multiple PKESKs if at least one does not use PQ/T encryption schemes.
 An implementation SHOULD NOT abort the encryption process when encrypting a message to both PQ/T and traditional keys to allow for a smooth transition to post-quantum cryptography.
 
-In case a certificate has both a PQ/T and an ECC encryption-capable valid subkey, the PQ/T subkey SHOULD be preferred.
 
 An implementation MAY sign with both a PQ(/T) and an ECC key using multiple signatures over the same data as described in {{multiple-signatures}}.
 Signing only with PQ(/T) key material is not backwards compatible.
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -894,7 +894,8 @@ Therefore various migration considerations must be taken into account, in partic
 ## Key preference {#pq-key-preference}
 
 Implementations SHOULD prefer PQ(/T) keys when multiple options are available.
-For instance, an implementation that encrypts messages for a recipient for which both a valid PQ/T and a valid ECC certificate are available, the implementation should use the PQ/T certificate only.
+In case that a certificate has both a valid PQ/T and a valid traditional subkey, an implementation SHOULD use the PQ/T subkey only.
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -321,9 +321,8 @@ Furthermore, multiple OpenPGP signatures may be combined on the application laye
 These latter two cases realize non-composite combinations of signatures.
 {{multiple-signatures}} specifies how implementations should handle the verification of such combinations of signatures.
 
-Furthermore, the OpenPGP protocol also allows for parallel encryption to different keys held by the same recipient.
-Accordingly, if the sender makes use of this feature and sends an encrypted message with multiple PKESK packages for different encryption keys held by the same recipient, a non-composite multi-algorithm public-key encryption is realized where the recipient has to decrypt only one of the PKESK packages in order to decrypt the message.
-See {{no-pq-t-parallel-encryption}} for restrictions on parallel encryption mandated by this specification.
+Furthermore, the OpenPGP protocol also allows for parallel encryption to different keys.
+Accordingly, when a message is encrypted using multiple PKESK packages for different encryption keys, a non-composite multi-algorithm public-key encryption is realized where the communication party has to decrypt only one of the PKESK packages in order to decrypt the message.
```
falko-strenzke, line 325: This is not pertaining to the changes, but to the text as was before: Actually, whether the encryption has to be considered "multi algorithm" or not depends on whether the set of keys used belong to different algorithms. So  probably we should rather write: 
```suggestion
Accordingly, when a message is encrypted using multiple PKESK packages for different encryption keys, a non-composite, potentially multi-algorithm, public-key encryption is realized where the communication party has to decrypt only one of the PKESK packages in order to decrypt the message.
```

falko-strenzke (commented):
I think the PR achieves what was agreed on in the interim. I am not sure that the new term "communication party" is avoiding the notion of a message transmission completely, but it clearly seems to be an improvement over "recipient". I also don't find any better term.

Once my proposed correction was decided on, I will approve this PR.

draft-ietf-openpgp-pqc.md
```diff
@@ -310,7 +310,7 @@ This specification introduces new cryptographic schemes, which can be categorize
 
  - PQ digital signature, namely SLH-DSA-SHAKE as a standalone cryptographic algorithm.
 
-For each of the composite schemes, this specification mandates that the recipient has to successfully perform the cryptographic algorithms for each of the component schemes used in a cryptographic message, in order for the message to be deciphered and considered as valid.
```
TJ-91, line 313: Since we're talking about signatures here, I guess "verifier" or something like that would work.

An alternative would be to simply state how the verification procedure works, not how an entity (recipient/comm. party) has to perform it. Something like "composite sigs are only considered valid if all component sigs are valid"

"communication party" might be weird in some contexts.

draft-ietf-openpgp-pqc.md
```diff
@@ -870,7 +868,7 @@ The algorithm-specific part of the secret key consists of:
 ## Symmetric Algorithms for SEIPD Packets
 
 Implementations MUST implement `AES-256`.
-An implementation SHOULD use `AES-256` in the case of a v1 SEIPD packet, or `AES-256` with any available AEAD mode in the case of a v2 SEIPD packet, if all recipients indicate support for it (explicitly or implicitly).
+An implementation SHOULD use `AES-256` in the case of a v1 SEIPD packet, or `AES-256` with any available AEAD mode in the case of a v2 SEIPD packet, if all communication parties indicate support for it (explicitly or implicitly).
```
TJ-91, line 871: I think the feedback in the Interim was not to avoid the word "recipient" everywhere. I think here it is sufficiently clear what is meant by "recipient" and I think it's the best word to use. The ambiguity comes in when matching multiple keys/certs to a recipient / entity and it becomes a more abstract concept.

For a PKESK packet there is a recipient / recipient key and for a SEIPD message the recipients are each recipient of the list of PKESK packets. I think that's common wording also in the Crypto Refresh.

falko-strenzke: Yes, it is correct the the term "recipient" is used throughout the crypto-refresh. So we should rather stick to it and probably just avoid conveying the idea of differentiating between keys that belong to a single recipient and to multiple recipients.

draft-ietf-openpgp-pqc.md
```diff
@@ -894,19 +892,19 @@ The post-quantum KEM algorithms defined in {{kem-alg-specs}} and the signature a
 During the transition period, the post-quantum algorithms will not be supported by all clients.
 Therefore various migration considerations must be taken into account, in particular backwards compatibility to existing implementations that have not yet been updated to support the post-quantum algorithms.
 
-## Key preference
+## Key preference {#pq-key-preference}
 
 Implementations SHOULD prefer PQ(/T) keys when multiple options are available.
+For instance, an implementation that encrypts messages for a recipient for which both a valid PQ/T and a valid ECC certificate are available, the implementation should use the PQ/T certificate only.
```
TJ-91, line 898: Here I think we run into the problem that the abstract concept of a recipient (outside of PKESKs) is not well defined. Might be ok for the non-normative statement, though.

draft-ietf-openpgp-pqc.md
```diff
@@ -894,19 +892,19 @@ The post-quantum KEM algorithms defined in {{kem-alg-specs}} and the signature a
...
+As specified in {{composite-kem}}, the confidentiality of a message is not post-quantum secure when using multiple PKESKs if at least one does not use PQ/T encryption schemes.
+An implementation SHOULD NOT abort the encryption process when encrypting a message to both PQ/T and traditional keys to allow for a smooth transition to post-quantum cryptography.
 
-For instance, if encrypting for a recipient for which both a valid PQ/T and a valid ECC certificate are available, the implementation SHOULD choose the PQ/T certificate.
 In case a certificate has both a PQ/T and an ECC encryption-capable valid subkey, the PQ/T subkey SHOULD be preferred.
 
 An implementation MAY sign with both a PQ(/T) and an ECC key using multiple signatures over the same data as described in {{multiple-signatures}}.
 Signing only with PQ(/T) key material is not backwards compatible.
 
-Note that the confidentiality of a message is not post-quantum secure when encrypting to multiple recipients if at least one recipient does not support PQ/T encryption schemes.
```
TJ-91, line 907: I think that here the word "recipient" would be unterstood in the context of the PKESK packet(s) so it'd be ok to use

draft-ietf-openpgp-pqc.md
```diff
@@ -997,7 +995,7 @@ This specification mandates support for `AES-256` for two reasons.
 First, `AES-KeyWrap` with `AES-256` is already part of the composite KEM construction.
 Second, some of the PQ(/T) algorithms target the security level of `AES-256`.
 
-For the same reasons, this specification further recommends the use of `AES-256` if it is supported by all recipients, regardless of what the implementation would otherwise choose based on the recipients' preferences.
+For the same reasons, this specification further recommends the use of `AES-256` if it is supported by all communication parties, regardless of what the implementation would otherwise choose based on the communication parties' preferences.
```
TJ-91, line 998: Also here I think "recipient" is ok to use

draft-ietf-openpgp-pqc.md
wussler: I don't think it's verification only here, if we want to use `verifier` we also need to say `decrypter`?

draft-ietf-openpgp-pqc.md
wussler:
```suggestion
An implementation SHOULD use `AES-256` in the case of a v1 SEIPD packet, or `AES-256` with any available AEAD mode in the case of a v2 SEIPD packet, if all recipients indicate support for it (explicitly or implicitly).
```

draft-ietf-openpgp-pqc.md
wussler: I think it's fine as an explaination of the previous sentence

draft-ietf-openpgp-pqc.md
dkg: seems to me that we have a "generating" implementation and a "consuming" or "evaluating" implementation.  "Communication party" is rather too vague.

draft-ietf-openpgp-pqc.md
```diff
@@ -321,9 +321,8 @@ Furthermore, multiple OpenPGP signatures may be combined on the application laye
 These latter two cases realize non-composite combinations of signatures.
 {{multiple-signatures}} specifies how implementations should handle the verification of such combinations of signatures.
 
-Furthermore, the OpenPGP protocol also allows for parallel encryption to different keys held by the same recipient.
-Accordingly, if the sender makes use of this feature and sends an encrypted message with multiple PKESK packages for different encryption keys held by the same recipient, a non-composite multi-algorithm public-key encryption is realized where the recipient has to decrypt only one of the PKESK packages in order to decrypt the message.
-See {{no-pq-t-parallel-encryption}} for restrictions on parallel encryption mandated by this specification.
+Furthermore, the OpenPGP protocol also allows for parallel encryption to different keys.
+Accordingly, when a message is encrypted using multiple PKESK packages for different encryption keys, a non-composite, potentially multi-algorithm, public-key encryption is realized where the communication party has to decrypt only one of the PKESK packages in order to decrypt the message.
```
dkg, line 325: what is a "PKESK package"?  does this mean "packet"?  i recognize it's not a part of the text edited here, but it could be cleaned up while we're touching this line.

again, i think "communication party" is too generic.  we're talking about the "consuming" or "evaluating" implementation, not the generating implementation.  in this case, we're just talking about decryption, so we could just say "the decrypting party".

dkg (dismissed):
I'm with @TJ-91 in that i think "communication party" has gotten too abstract to be useful.  can we pull back a little bit?

draft-ietf-openpgp-pqc.md
```diff
@@ -363,15 +362,14 @@ This draft will not be sent to IANA without every listed algorithm having a non-
...
 The ML-KEM + ECDH public-key encryption involves both the ML-KEM and an ECDH KEM in an a priori non-separable manner.
 This is achieved via KEM combination, i.e. both key encapsulations/decapsulations are performed in parallel, and the resulting key shares are fed into a key combiner to produce a single shared secret for message encryption.
 
-## Parallel Public-Key Encryption {#no-pq-t-parallel-encryption}
-
-As explained in {{non-composite-multi-alg}}, the OpenPGP protocol inherently supports parallel encryption to different keys of the same recipient.
-Implementations MUST NOT encrypt a message with a purely traditional public-key encryption key of a recipient if it is encrypted with a PQ/T key of the same recipient.
+As explained in {{non-composite-multi-alg}}, the OpenPGP protocol inherently supports parallel encryption to different keys.
+Note that the confidentiality of a message is not post-quantum secure when encrypting to different keys if at least one key does not support PQ/T encryption schemes.
+In section {{pq-key-preference}} is explained how to deal with multiple key scenarios.
```
falko-strenzke, line 372:
```suggestion
In section {{pq-key-preference}} it is explained how to deal with multiple key scenarios.
```

draft-ietf-openpgp-pqc.md
```diff
@@ -929,7 +927,7 @@ Two migration strategies are recommended:
 ## Security Aspects of Composite Signatures
 
 When multiple signatures are applied to a message, the question of the protocol's resistance against signature stripping attacks naturally arises.
-In a signature stripping attack, an adversary removes one or more of the transmitted signatures such that only a subset of the signatures originally applied by the sender remain in the message that reaches the recipient.
+In a signature stripping attack, an adversary removes one or more of the signatures such that only a subset of the signatures remain in the message when verified.
```
falko-strenzke, line 930:
```suggestion
In a signature stripping attack, an adversary removes one or more of the signatures such that only a subset of the signatures remain in the message at the point when it is verified.
```

falko-strenzke (approved):
Looks good to me. Thanks for the work Aron. I just propose two corrections regarding wording.

draft-ietf-openpgp-pqc.md
TJ-91: I would still rather rewrite this. IMO that problem is outside of the scope of the OpenPGP protocol and more an application thing. The protocol layer doesn't necessarily know that two certificates / keys that it processes for encrypting are somehow linked. And even if it does, it can't know the intent of the application/user.
See also #67, I think the critique still applies to this: Do we mean the second or the third scenario, or both, and what is the criterion for an application to determine that two certificates belong to the same recipient in the third scenario?

In any case, I would change the paragraph to start with the more obvious case first:

> For instance, in the case that a certificate has both a PQ/T and an ECC encryption-capable valid subkey, the PQ/T subkey SHOULD be preferred.

And then we can add the "multiple certificate scenario" a bit more vaguely. The best I could come up with is the following:

> Further, if an application can determine (by any means not defined in this document) that encrypting to a PQ/T certificate and a traditional certificate serves the same purpose, it should omit encrypting to the traditional certificate.

Sorry for being a bit picky here but I think sooner or later someone will re-raise this issue.

draft-ietf-openpgp-pqc.md
```diff
@@ -321,9 +321,8 @@ Furthermore, multiple OpenPGP signatures may be combined on the application laye
 These latter two cases realize non-composite combinations of signatures.
 {{multiple-signatures}} specifies how implementations should handle the verification of such combinations of signatures.
 
-Furthermore, the OpenPGP protocol also allows for parallel encryption to different keys held by the same recipient.
-Accordingly, if the sender makes use of this feature and sends an encrypted message with multiple PKESK packages for different encryption keys held by the same recipient, a non-composite multi-algorithm public-key encryption is realized where the recipient has to decrypt only one of the PKESK packages in order to decrypt the message.
-See {{no-pq-t-parallel-encryption}} for restrictions on parallel encryption mandated by this specification.
+Furthermore, the OpenPGP protocol also allows for parallel encryption to different keys.
+Accordingly, when a message is encrypted using multiple PKESK packets for different encryption keys, a non-composite, potentially multi-algorithm, public-key encryption is realized where the consuming party has to decrypt only one of the PKESK packets in order to decrypt the message.
```
TJ-91, line 325: I'd try to state this more concisely since it's more or less stating the obvious.

```suggestion
Furthermore, the OpenPGP protocol also allows parallel encryption to different keys by using multiple PKESK packets, thus realizing non-composite multi-algorithm public-key encryption.
```

draft-ietf-openpgp-pqc.md
falko-strenzke: After looking closer again I agree with Johannes, here the assumptions about the identity of the recipient will most likely have to be more vague.

wussler:
```suggestion
For instance, in the case that a certificate has both a PQ/T and an ECC encryption-capable valid subkey, the PQ/T subkey SHOULD be preferred.
Furthermore, if an application can anyhow determine that encrypting to a PQ/T certificate and a traditional certificate is redundant, it should omit encrypting to the traditional certificate.
```
What about this?

falko-strenzke: I think we're almost there. But "SHOULD be preferred" might not be clear enough, as we also want to express that in this case the traditional key should not be used at all. So I would write for instance " ... valid subkey, the PQ/T subkey SHOULD be used and the traditional key SHOULD NOT be used."

Furthermore, I wouldn't write "and an *ECC* encryption-capable valid subkey" but rather "and a *traditional* encryption-capable valid subkey" in order not to create the false assumption that it only refers to ECC keys (other types of keys might be around for quite some time).

wussler: That's the point of "for instance", IMO. I would rather remove the normative part from an example then or remove the "for instance".

Normative way:
```suggestion
In case that a certificate has both a valid PQ/T and a valid traditional subkey, an implementation SHOULD use the PQ/T subkey only.
Furthermore, if an application can anyhow determine that encrypting to a PQ/T certificate and a traditional certificate is redundant, it should omit encrypting to the traditional certificate.
```
Example way:
```suggestion
For instance, an implementation that encrypts messages for a recipient for which both a valid PQ/T and a valid ECC encryption subkey are available, the implementation should use the PQ/T subkey only.
Furthermore, if an application can anyhow determine that encrypting to a PQ/T certificate and a traditional certificate is redundant, it should omit encrypting to the traditional certificate.
```
Somehow hybrid, but not normative:
```suggestion
For instance, an implementation that encrypts messages for a recipient for which both a valid PQ/T and a valid traditional encryption subkey are available, the implementation should use the PQ/T subkey only.
Furthermore, if an application can anyhow determine that encrypting to a PQ/T certificate and a traditional certificate is redundant, it should omit encrypting to the traditional certificate.
```

TJ-91: I would write:

> For instance, an implementation that encrypts a message to a certificate for which both a valid PQ/T and a valid traditional encryption subkey are available, the implementation should encrypt to the PQ/T subkey only.

This is adapted from your "Example way" to avoid the "recipient" terminology, and I also replaced "use" by "encrypt to", as well as "ECC" by "traditional".

I'm happy with your suggestion for the "multiple certificate scenario".

falko-strenzke: I vote for the normative variant @wussler has posted above.

(An example with ECC is in my point potentially misleaing – even though from a formal point it is correct since it is an example. But our aim here by rendering the guidance as an example is only with respect to the set of keys, I don't see a need to make it concrete with respect to the public-key encryption algorithm.)

I also generally agree to the second sentence about the redundant certificate. That solves the problem of the using the word recipient well in my view. But the use of the word "anyhow" doesn't seem right to me. I ran the sentence through deepl and it doesn't translate to German correctly. I think we need to replace "anyhow" with "in some way" or with "somehow". Or maybe even better 

"Furthermore, if an application has any means to determine ..."

TJ-91: Ok, then let's go with the normative variant.

I was a bit unsure about "anyhow" myself but cannot judge it. I vote for your suggestion to write "Furthermore, if an application has any means to determine ...".

falko-strenzke (approved):

draft-ietf-openpgp-pqc.md
```diff
@@ -894,19 +891,20 @@ The post-quantum KEM algorithms defined in {{kem-alg-specs}} and the signature a
 During the transition period, the post-quantum algorithms will not be supported by all clients.
 Therefore various migration considerations must be taken into account, in particular backwards compatibility to existing implementations that have not yet been updated to support the post-quantum algorithms.
 
-## Key preference
+## Key preference {#pq-key-preference}
 
 Implementations SHOULD prefer PQ(/T) keys when multiple options are available.
+In case that a certificate has both a valid PQ/T and a valid traditional subkey, an implementation SHOULD use the PQ/T subkey only.
```
TJ-91, line 897: Sorry for the holdup, but I just noticed that we should phrase it a bit narrower. Otherwise this statement also includes the signature case and contradicts what we write in line 905, 906.

```suggestion
When encrypting to a certificate that has both a valid PQ/T and a valid traditional encryption subkey, an implementation SHOULD use the PQ/T subkey only.
```

draft-ietf-openpgp-pqc.md
```diff
@@ -894,19 +891,20 @@ The post-quantum KEM algorithms defined in {{kem-alg-specs}} and the signature a
...
+Furthermore, if an application has any means to determine that encrypting to a PQ/T certificate and a traditional certificate is redundant, it should omit encrypting to the traditional certificate.
+
+As specified in {{composite-kem}}, the confidentiality of a message is not post-quantum secure when using multiple PKESKs if at least one does not use PQ/T encryption schemes.
+An implementation SHOULD NOT abort the encryption process when encrypting a message to both PQ/T and traditional keys to allow for a smooth transition to post-quantum cryptography.
 
-For instance, if encrypting for a recipient for which both a valid PQ/T and a valid ECC certificate are available, the implementation SHOULD choose the PQ/T certificate.
 In case a certificate has both a PQ/T and an ECC encryption-capable valid subkey, the PQ/T subkey SHOULD be preferred.
```
TJ-91, line 903: still needs to be removed

TJ-91, line 903:
```suggestion
```

TJ-91 (approved):

Issue #61, opened 2023-09-01
Format hyperlinks to mentioned sections in external documents (e.g. RFCs)


fluppe2:
Makes dev life easier.

fluppe2:
Done in #136.

PR #136, opened 2024-08-23
Updated refs for crypto-refresh to point to RFC9580. Formatted refere…

fluppe2:
…nces to sections in external RFCs to be direct hyperlinks.

This adresses #134 as well as #61.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -50,7 +50,7 @@ normative:
 
   RFC3394:
 
-  I-D.ietf-openpgp-crypto-refresh:
+  RFC9580:
 
 informative:
 
@@ -253,7 +253,7 @@ The document specifies the conventions for interoperability between compliant Op
 
 ### Terminology for Multi-Algorithm Schemes
 
-The terminology in this document is oriented towards the definitions in {{?I-D.ietf-pquip-pqt-hybrid-terminology}}.
+The terminology in this document is oriented towards the definitions in [RFC9580].
 Specifically, the terms "multi-algorithm", "composite" and "non-composite" are used in correspondence with the definitions therein.
 The abbreviation "PQ" is used for post-quantum schemes.
 To denote the combination of post-quantum and traditional schemes, the abbreviation "PQ/T" is used.
@@ -316,7 +316,7 @@ In the case of the composite public-key decryption, each of the component KEM de
 
 ### Non-Composite Algorithm Combinations {#non-composite-multi-alg}
 
-As the OpenPGP protocol [I-D.ietf-openpgp-crypto-refresh] allows for multiple signatures to be applied to a single message, it is also possible to realize non-composite combinations of signatures.
+As the OpenPGP protocol [RFC9580] allows for multiple signatures to be applied to a single message, it is also possible to realize non-composite combinations of signatures.
 Furthermore, multiple OpenPGP signatures may be combined on the application layer.
 These latter two cases realize non-composite combinations of signatures.
 {{multiple-signatures}} specifies how implementations should handle the verification of such combinations of signatures.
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -253,7 +253,7 @@ The document specifies the conventions for interoperability between compliant Op
 
 ### Terminology for Multi-Algorithm Schemes
 
-The terminology in this document is oriented towards the definitions in {{?I-D.ietf-pquip-pqt-hybrid-terminology}}.
+The terminology in this document is oriented towards the definitions in [RFC9580].
```
falko-strenzke, line 256: This was a different reference.
```suggestion
The terminology in this document is oriented towards the definitions in {{?I-D.ietf-pquip-pqt-hybrid-terminology}}.
```

draft-ietf-openpgp-pqc.md
```diff
@@ -379,13 +379,13 @@ The ML-DSA + EdDSA signature consists of independent ML-DSA and EdDSA signatures
 The OpenPGP message format allows multiple signatures of a message, i.e. the attachment of multiple signature packets.
 
 An implementation MAY sign a message with a traditional key and a PQ(/T) key from the same sender.
-This ensures backwards compatibility due to {{I-D.ietf-openpgp-crypto-refresh}}, Section 5.2.5, since a legacy implementation without PQ(/T) support can fall back on the traditional signature.
+This ensures backwards compatibility due to [[RFC9580, Section 5.2.5]](https://www.rfc-editor.org/rfc/rfc9580#name-malformed-and-unknown-signa), since a legacy implementation without PQ(/T) support can fall back on the traditional signature.
```
falko-strenzke, line 382: I am not entirely sure whether this way of rendering the reference is common practice. In the new rendering, there is no more a link to the reference section in the same document. This means that the references to RFC 9580 without a specific section point to the reference section, and the references to the RFC with a specific section point to the URL.  Does anyone know an example of an RFC which uses references in this way?

According to https://www.rfc-editor.org/styleguide/part2/ the new variant seems to be OK, as the prescription there only pertains to the rendering of the references-section, and not to how the references in the text have to look.

@TJ-91 @wussler Any opinions on this?

falko-strenzke (changes requested):

draft-ietf-openpgp-pqc.md
wussler: Maybe I am being dense, but do not see the problem with this reference format

wussler (approved):
LGTM, if applied the comment from @falko-strenzke (terminology ref)

draft-ietf-openpgp-pqc.md
falko-strenzke: Yeah, I think it'll be OK. It's just that then you don't jump to the reference section by following the link, which is the only format I have seen so far in RFCs. But in any case, if our format should turn out not be OK, someone will tell us at some point and we can still revert it. In any case it is more convenient to jump to the referenced section directly.

TJ-91 (commented):
Note that you can also use the section number in the URL: https://www.rfc-editor.org/rfc/rfc9580#section-5.2.5 instead of https://www.rfc-editor.org/rfc/rfc9580#name-malformed-and-unknown-signa

This might be easier to maintain and spot errors.

draft-ietf-openpgp-pqc.md
fluppe2: Thank you for spotting this.

draft-ietf-openpgp-pqc.md
fluppe2: Ok, seems like I can resolve this conversation.

fluppe2:
> Note that you can also use the section number in the URL: https://www.rfc-editor.org/rfc/rfc9580#section-5.2.5 instead of https://www.rfc-editor.org/rfc/rfc9580#name-malformed-and-unknown-signa
> 
> This might be easier to maintain and spot errors.

Applied, thank you.

falko-strenzke (approved):

PR #145, opened 2024-10-10
Pure EdDSA, ML-DSA, SLH-DSA. Update refs to final FIPS 203, 204, 205

### LINKED ISSUE #126: Mention that no further hashing happens in EdDSA
We will need to refer to this section in the crypto-refresh: https://www.rfc-editor.org/rfc/rfc9580.html#eddsa-notes

It specifies exactly how EdDSA is invoked.

@wussler @TJ-91

fluppe2:
This should also close #126.

I think the sentences added to the specific sections should be enough.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -171,25 +171,25 @@ informative:
       NIST Special Publication 800-108 Rev. 1
 
   FIPS-203:
-      target: https://doi.org/10.6028/NIST.FIPS.203.ipd
+      target: https://doi.org/10.6028/NIST.FIPS.203
       title: Module-Lattice-Based Key-Encapsulation Mechanism Standard
       author:
         - org: National Institute of Standards and Technology
-      date: August 2023
+      date: August 2024
 
   FIPS-204:
-      target: https://doi.org/10.6028/NIST.FIPS.204.ipd
+      target: https://doi.org/10.6028/NIST.FIPS.204
       title: Module-Lattice-Based Digital Signature Standard
       author:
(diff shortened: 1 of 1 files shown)


fluppe2:
> The only thing I'm not sure about is whether or not "Pure ML-DSA" is well enough established in FIPS 204. The variant is only called "pure" one time and the "pure" is put in quotation marks. FIPS 205 is much more clear on that.

Good point, but remember that we are explicitly referencing the algorithms 'ML-DSA.Sign' and 'ML-DSA.Verify' which constitute the pure version of ML-DSA. We are not pointing to 'HashML-DSA.Sign/Verify'. Do you think that is fine enough?

TJ-91:
> we are explicitly referencing the algorithms 'ML-DSA.Sign' and 'ML-DSA.Verify' 

That's true. I think it's fine.

draft-ietf-openpgp-pqc.md
```diff
@@ -717,6 +720,9 @@ TBD (108 for testing)  | Ed448   | 57         | 57         | 57         | 114
 
 ### ML-DSA signatures {#mldsa-signature}
 
+Throughout this specification ML-DSA refers to the pure ML-DSA version defined
```
falko-strenzke, line 723:
```suggestion
Throughout this specification ML-DSA refers to the pure version ML-DSA, i.e., in contrast to the pre-hash variant, defined
```
I think this clarification is necessary as unfortunately, as Johannes pointed out, the label "pure" is not formally defined FIPS-204.

wussler (approved):

PR #146, opened 2024-10-16
Defining ML-KEM and ML-DSA secret keys to be in seed format

### LINKED ISSUE #108: Specify private key format for ML-KEM and ML-DSA when switching to FIPS final versions
Based on what was announced, NIST' s final standard for the ML-KEM and ML-DSA will enable private keys in seed format. Currently, the draft only refers to the NIST standard for the private key format. Thus this needs more specification, and it might be relevant to specify in the draft:

- what is the single (?) secret key format in the NIST standard, that the draft is referring to
  - alternative option: that seed format is also possible for private keys (?)
- that a private key in expanded format needs to be checked for consistency before usage

### LINKED ISSUE #139: Private key storage in seed vs expanded key format for ML-KEM, ML-DSA
As I raised on the mailing list [1], I think there's a reasonable argument for limiting the storage format for post-quantum keys to the seed format rather than the, arguably more temperamental, expanded key format.  I think LAMPS is going in the direction of seed-format-only so I think there is precendence that we can follow and go with the same approach. See [2] and [3]

There were some concerns about low-power devices/OpenPGP smart cards but I think it was decided that was not a huge concern or at least not a reason to have multiple key storage formats (but others are free to contradict me if it is :) ).

Happy to provide some assistance with writing/reviewing text on storage formats if it'll help.

[1] https://mailarchive.ietf.org/arch/msg/openpgp/2dg-DrWt4DgIoX8uTalCa-kzE38/
[2] https://github.com/randombit/botan/pull/3893#issuecomment-2349229966
[3] https://github.com/randombit/botan/pull/4270#issuecomment-2356039856

falko-strenzke:
Closes #139 
Closes #108

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -519,8 +519,8 @@ All artifacts are encoded as defined in [FIPS-203].
 {: title="ML-KEM parameters artifact lengths in octets" #tab-mlkem-artifacts}
 Algorithm ID reference | ML-KEM      | Public key | Secret key | Ciphertext | Key share
 ----------------------:| ----------- | ---------- | ---------- | ---------- | ---------
-TBD (105 for testing)  | ML-KEM-768  | 1184       | 2400       | 1088       | 32
-TBD (106 for testing)  | ML-KEM-1024 | 1568       | 3168       | 1568       | 32
+TBD (105 for testing)  | ML-KEM-768  | 1184       | 64         | 1088       | 32
+TBD (106 for testing)  | ML-KEM-1024 | 1568       | 64         | 1568       | 32
 
 To instantiate `ML-KEM`, one must select a parameter set from the column "ML-KEM" of {{tab-mlkem-artifacts}}.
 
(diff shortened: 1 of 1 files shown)


TJ-91 (commented):
Looks good to me.

One further issue is that we use `ML-KEM.Decaps` and `ML-DSA.Sign` directly with the secret key (seed) which is only defined for the expanded key. For ML-KEM we would first need to call `ML-KEM.KeyGen_internal` to obtain the actual key that can be input to Decaps.

falko-strenzke:
> We use `ML-KEM.Decaps` and `ML-DSA.Sign` directly with the secret key (seed) which is only defined for the expanded key. For ML-KEM we would first need to call `ML-KEM.KeyGen_internal` to obtain the actual key that can be input to Decaps.

This is why it says in the added text to the secret key format:

> Upon parsing the private key format, or before using the secret key, for the expansion of the key, the function `ML-KEM.KeyGen_internal` [FIPS-203] has to be invoked [...]

I think it is easier to state that where the key format is defined, because it is up to the implementation when to expand the secret key.

TJ-91 (approved):
I re-read and now agree with you. It should be sufficiently clear, especially since we clarify that `mlkemSecretKey` is the decapsulation key.

fluppe2 (approved):
LGTM

wussler (approved):
Same for me, with changelog looks good!

TJ-91 (approved):

PR #157, opened 2024-11-08
assign code points 30 - 34 for ML-DSA and SLH-DSA

TJ-91:
I simply assign code points for the signature algorithms starting with 30 (since 29 is reserved due to LibrePGP).

I can add updated test vectors to this PR but would suggest waiting for possible KDF changes first.


falko-strenzke (approved):

wussler (approved):
Looks good to me. Missing changelog entry

PR #159, opened 2024-11-08
Aligning KEM Combiner with LAMPS

ounsworth:


draft-ietf-openpgp-pqc.md
```diff
@@ -567,14 +565,13 @@ It is given by the following algorithm, which computes the key encryption key `K
     //  domSep given in hexadecimal encoding := 4F 70 65 6E 50 47 50 43 6F 6D 70
     //                                          6F 73 69 74 65 4B 44 46 76 31
 
-    KEK = KMAC256(mlkemKeyShare || ecdhKeyShare, mlkemCipherText || ecdhCipherText
-                  || mlkemPublicKey || ecdhPublicKey || algId, 256, domSep)
+    KEK = SHA3-256(mlkemKeyShare || ecdhKeyShare || ecdhCipherText || ecdhPublicKey || domSep)
```
falko-strenzke, line 568: `algId` should be added at the end.

fluppe2:
This PR will need to

- substantially revise the security considerations for the KEM combiner. They are not adequate if the ML-KEM ciphertext is missing, as the present considerations build upon a modular construction.

- include a warning in the security considerations, that this KEM combiner builds upon specific security properties of ML-KEM and as such is tailored to ML-KEM. That is, extreme care must be taken when another PQ-KEM is discussed (possibly in the future).

- the KEM combiner (with missing ML-KEM ciphertext) does not scale beyond 128 Bits of security (due to the birthday bound arguments in the X-Wing paper). Alternatively give scientifically sound arguments why it indeed scales.

These aspects must be adressed before we merge this.

draft-ietf-openpgp-pqc.md
```diff
@@ -567,14 +565,13 @@ It is given by the following algorithm, which computes the key encryption key `K
...
 
-Note that the values `ecdhKeyShare` defined in {{ecc-kem}} and `mlkemKeyShare` defined in {{mlkem-ops}} already use the relative ciphertext in the derivation.
-The ciphertext and public keys are by design included again in the key combiner to provide a robust security proof.
+Note that the values `ecdhKeyShare` defined in {{ecc-kem}} already uses the relative ciphertext in the derivation.
+The `ecdhCipherText` and `ecdhPublicKey` are by design included again in the key combiner to provide a robust security proof.
+The ML-KEM algorithm already explicitly binds the ML-KEM public key in `ML-KEM.Encaps` (Algorithm 17, line 1 in [FIPS-203]), and implicitly binds the ML-KEM public key and ciphertext in the Fujisaki-Okamoto confirm step of `ML-KEM.Decaps` (Algorithm 18, line 9 in [FIPS-203]) and therefore they are not included in this KEM combiner construction. Due to this, this KEM combiner should be considered to be optimized for ML-KEM and that it does not generatize to be secure over arbitrary other KEMs.
```
fluppe2, line 574:
```suggestion
The ML-KEM algorithm already explicitly binds the ML-KEM public key in `ML-KEM.Encaps` (Algorithm 17, line 1 in [FIPS-203]), and implicitly binds the ML-KEM public key and ciphertext in the Fujisaki-Okamoto confirm step of `ML-KEM.Decaps` (Algorithm 18, line 9 in [FIPS-203]) and therefore they are not included in this KEM combiner construction. Due to this, this KEM combiner should be considered to be optimized for ML-KEM and that it does not generalize to be secure over arbitrary other KEMs.
```

fluppe2 (changes requested):
This PR will need to
- revise the CCA security considerations for the KEM combiner. They are not adequate if the ML-KEM ciphertext is missing, as the present considerations build upon a modular construction.
- repeat the warning in the security considerations, that this KEM combiner builds upon specific security properties of ML-KEM and as such is tailored to ML-KEM. That is, extreme care must be taken when another PQ-KEM is discussed (possibly in the future).
- state that the KEM combiner (with missing ML-KEM ciphertext) does not scale beyond 128 Bits of security (due to the birthday bound arguments in the X-Wing paper). Alternatively give scientifically sound arguments why it indeed scales.

These aspects must be adressed before we proceed to a merge.

falko-strenzke:
> This PR will need to
> 
>     * revise the CCA security considerations for the KEM combiner. They are not adequate if the ML-KEM ciphertext is missing, as the present considerations build upon a modular construction.
> 
>     * repeat the proposed warning (in the security considerations), that this KEM combiner builds upon specific security properties of ML-KEM and as such is tailored to ML-KEM. That is, extreme care must be taken when another PQ-KEM is discussed (possibly in the future).
> 
>     * state that the KEM combiner (with missing ML-KEM ciphertext) does not scale beyond 128 Bits of security (due to the birthday bound arguments in the X-Wing paper). Alternatively give scientifically sound arguments why it indeed scales.
> 
> 
> These aspects must be adressed before we proceed to a merge.

Thanks for pointing out theses aspects, Stavros. Generally I wouldn't proceed directly to merging it. This topic was discussed before in the WG and there at least one implementer was clearly requesting a generic construction that could be reused also for other KEMs. Before we introduce a combiner that doesn't fulfill this requirement, I think we need to inform the list first. Not all WG members might see exactly why OpenPGP has to per se align to LAMPS (even though clearly there was a number of voices requesting that recently, also in the meeting). That being said, I am not against going down that road, but I think it requires a confirmation on the list first. Otherwise we risk that we put a new version into the draft that then is again subject to discussion.

Still, I think completing this PR might make sense in order to have a concrete proposal to point to when bringing it to the list.

teythoon:
Hi Mike, thanks for proposing the concrete change.

For context, this came out of a bit of a hallway discussion after the IETF 121 working group meeting.  The motivation is two-fold:

First, don't get hung up on the FIPS compliance question regarding the combiner.  It seems to me that lamps found a compliant combiner, whereas our current proposal has some issues or at least there are open questions regarding where the key shares should go (and indeed this might reverse in the future as FIPS evolves).

Second, for me as implementer, SHA3 is more easily available than KMAC.

I know that I pushed for having a single combiner applicable to all KEMs.  I'm happy to relax this requirement, i.e. I'd be fine if you say that for ML-KEM, use this combiner, and whenever we introduce a new KEM, it might bring its own combiner.

wussler:
Superseded by !161

PR #161, opened 2024-11-28
Align KEM Combiner and drop ECDH-KEM CCA-conversion

fluppe2:
- Modified KEM Combiner
- Dropped ECDH-KEM CCA-conversion
- Aligned security considerations
- Removed and added references appropriately
- Added Changelog under -07

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -613,7 +613,7 @@ It is given by the following algorithm, which computes the key encryption key `K
     return KEK
 
 Note that this is in line with {{I-D.ietf-lamps-pq-composite-kem}} up to `mlkemCipherText || mlkemPublicKey || algId || domSep` being
-the OpenPGP specific domain separation.
+the OpenPGP-specific domain separation.
 
 ### Key generation procedure {#ecc-mlkem-generation}
 
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -590,7 +590,7 @@ It is given by the following algorithm, which computes the key encryption key `K
     //   mlkemPublicKey  - the ML-KEM public key of the recipient as an octet string
     //   ecdhKeyShare    - the ECDH key share encoded as an octet string
     //   ecdhCipherText  - the ECDH ciphertext encoded as an octet string
-    //   ecdhPublicKey   - The ECDH public key of the recipient as an octet string
+    //   ecdhPublicKey   - the ECDH public key of the recipient as an octet string
     //   algId           - the OpenPGP algorithm ID of the public-key encryption algorithm
 
     KEK = SHA3-256( mlkemKeyShare || ecdhKeyShare || ecdhCipherText || ecdhPublicKey
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -594,15 +594,14 @@ It is given by the following algorithm, which computes the key encryption key `K
     //   algId           - the OpenPGP algorithm ID of the public-key encryption algorithm
 
     KEK = SHA3-256( mlkemKeyShare || ecdhKeyShare || ecdhCipherText || ecdhPublicKey
-                    || mlkemCipherText || mlkemPublicKey || algId || const )
+                    || mlkemCipherText || mlkemPublicKey || algId || domSep )
     return KEK
 
-The value `const` is a constant set to the UTF-8 encoding of the string "OpenPGPCompositeKDFv1", i.e.
+The value `domSep` is a constant set to the UTF-8 encoding of the string "OpenPGPCompositeKDFv1", i.e.
 
-    const := 4F 70 65 6E 50 47 50 43 6F 6D 70 6F 73 69 74 65 4B 44 46 76 31
+    domSep := 4F 70 65 6E 50 47 50 43 6F 6D 70 6F 73 69 74 65 4B 44 46 76 31
 
-Note that this is in line with {{I-D.ietf-lamps-pq-composite-kem}} up to `mlkemCipherText || mlkemPublicKey || algId || const` being
-the OpenPGP-specific domain separation.
+Note that this is in line with {{I-D.ietf-lamps-pq-composite-kem}} up to `mlkemCipherText || mlkemPublicKey || algId || domSep` being the OpenPGP-specific context information.
 
 ### Key generation procedure {#ecc-mlkem-generation}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -565,14 +608,12 @@ It is given by the following algorithm, which computes the key encryption key `K
     //  domSep given in hexadecimal encoding := 4F 70 65 6E 50 47 50 43 6F 6D 70
```
TJ-91, line 608: Not directly related to your PR but in earlier versions we distinguished "Input" and "Constants" for `multiKeyCombine`. I'd be in favor of listing `domSep` as constant again.

draft-ietf-openpgp-pqc.md
```diff
@@ -565,14 +608,12 @@ It is given by the following algorithm, which computes the key encryption key `K
...
+    KEK = SHA3-256(mlkemKeyShare || ecdhKeyShare || ecdhCipherText || ecdhPublicKey
+                   || mlkemCipherText || mlkemPublicKey || algId || domSep)
     return KEK
 
-Here, the parameters to KMAC256 appear in the order as specified in {{SP800-185}}, Section 4, i.e., the key K, main input data X, requested output length in bits L, and optional customization string S.
-
-Note that the values `ecdhKeyShare` defined in {{ecc-kem}} and `mlkemKeyShare` defined in {{mlkem-ops}} already use the relative ciphertext in the derivation.
-The ciphertext and public keys are by design included again in the key combiner to provide a robust security proof.
+Note that this is in line with {{I-D.ietf-lamps-pq-composite-kem}} up to `mlkemCipherText || mlkemPublicKey || algId || domSep` being
+the OpenPGP specific domain separation.
```
TJ-91, line 616:
```suggestion
the OpenPGP-specific domain separation.
```

I think a hyphen is needed here

draft-ietf-openpgp-pqc.md
```diff
@@ -934,36 +975,32 @@ As a consequence, a component signature taken out of the context of a specific c
...
+    K = SHA3-256(K1 || K2 || C2 || P2 || C1 || P1 || algId || domSep)
 
 from a random bit string, an adversary has to know (or correctly guess) both key shares `K1` and `K2`, entirely.
 
 The proposed construction in {{kem-key-combiner}} preserves IND-CCA2 of any of its ingredient KEMs, i.e. the newly formed combined KEM is IND-CCA2 secure as long as at least one of the ingredient KEMs is.
 Indeed, the above stated indifferentiability from a random oracle qualifies Keccak as a split-key pseudorandom function as defined in {{GHP18}}.
 That is, Keccak behaves like a random function if at least one input shared secret is picked uniformly at random.
 Our construction can thus be seen as an instantiation of the IND-CCA2 preserving Example 3 in Figure 1 of {{GHP18}}, up to some reordering of input shared secrets and ciphertexts.
-In the random oracle setting, the reordering does not influence the arguments in {{GHP18}}.
+In the random oracle setting, the reordering does not influence the arguments in {{GHP18}}. That is, in the post-quantum setting the construction inherits its IND-CCA2 security from any ingredient post-quantum IND-CCA2 secure KEM.
```
TJ-91, line 997:
```suggestion
In the random oracle setting, the reordering does not influence the arguments in {{GHP18}}. That is, in the post-quantum setting the construction inherits its IND-CCA2 security from any ingredient post-quantum IND-CCA2-secure KEM.
```

I think a hyphen is needed here

TJ-91 (commented):
Looks good to me, only some minor editorial comments.

draft-ietf-openpgp-pqc.md
```diff
@@ -451,13 +494,13 @@ The operation `x25519Kem.Encaps()` is defined as follows:
 
  3. Set the output `ecdhCipherText` to `V`
 
- 4. Set the output `ecdhKeyShare` to `SHA3-256(X || ecdhCipherText || ecdhPublicKey)`
+ 4. Set the output `ecdhKeyShare` to `X`
```
falko-strenzke, line 497: We could simply say above
`ecdhKeyShare = X25519(v, R)`.
That saves one line in each algorithm description just to assign X. But maybe we want to keep `X` for clarity?

falko-strenzke (commented):
In principle it's fine in my judgement, we should just decide on / process the open comments.

wussler (approved):
The PR looks good to me. I'll follow up on the list with a link to this PR

draft-ietf-openpgp-pqc.md
fluppe2: I think I would prefer it as it is right now, the first steps describe the computational steps and the last "Set" statements take the role of a "return".

draft-ietf-openpgp-pqc.md
TJ-91: I think it's a good idea to rename it to avoid confusion. I'm fine with it being `const`. I can't think of a good alternative with a more descriptive name right now.

You forgot to rename `domSep` in 9.2. and 9.2.1.

ahuelsing (approved):
Looks good to me, although I personally would prefer PK instead of P as that may be read as plaintext by some.

falko-strenzke:
> Looks good to me, although I personally would prefer PK instead of P as that may be read as plaintext by some.

I agree with Andreas, it should better be "PK1" etc.

fluppe2:
> > Looks good to me, although I personally would prefer PK instead of P as that may be read as plaintext by some.
> 
> I agree with Andreas, it should better be "PK1" etc.

Good catch, see [88d3fd1](https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/161/commits/88d3fd148d44ac740e490a4ba6d4039ab41f6c0d).

draft-ietf-openpgp-pqc.md
fluppe2: I had an idea, see [55bea08](https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/161/commits/55bea08b450e49dd8cb8a2c9c401fe9e63a37cec).

That is:
- Stick to `domSep` (don't rename it to `const`)
- We refer to `mlkemCT || mlkemPK || algId || domSep` as the OpenPGP-specific **context** information
- Refactored text in Section 9.2.1 to say that the inclusion of the ciphertexts binds the KEK to the specific session

Do we need the last paragraph in Section 9.2.1 that says something about ECC and SP800-56A? It feels like a leftover from old discussions. I removed it, if you think we need it, include it again.

falko-strenzke: It's true that we discussed proposing to change the term "Domain" in the LAMPS draft. But when I think about it now, I fear that might delay the process. I would suggest to adopt to their terminology and find some other name for our string. Maybe "domsepString" or something similar?

draft-ietf-openpgp-pqc.md
```diff
@@ -934,43 +963,38 @@ As a consequence, a component signature taken out of the context of a specific c
...
-This ensures that the input keying material is used to generate a KEK for a specific purpose or context.
+The `domSep` information defined in {{kem-key-combiner}} provides the domain separation for the key combiner construction.
+This ensures that the input keying material is used to generate a KEK for a specific purpose.
 
-The algorithm ID, passed as the `algID` paramter to `multiKeyCombine`, binds the derived KEK to the chosen algorithm. The input of the public keys into `multiKeyCombine` binds the KEK to the communication parties.
+The algorithm ID, passed as the `algID` parameter to `multiKeyCombine`, binds the derived KEK to the chosen algorithm.
 The algorithm ID identifies unequivocally the algorithm, the parameters for its instantiation, and the length of all artifacts, including the derived key.
 
-This is in line with the Recommendation for ECC in Section 5.5 of [SP800-56A].
-Other fields included in the recommendation are not relevant for the OpenPGP protocol, since the sender is not required to have a key of their own, there are no pre-shared secrets, and all the other parameters are unequivocally defined by the algorithm ID.
```
falko-strenzke, line 973: I think the two previous lines were simly included to show that we are using sufficient context according to SP800-56A. I agree that we don't necessarily need them.

draft-ietf-openpgp-pqc.md
fluppe2: Why not leave our version as it is right now with this terminology. It is a only a proposal anyway, and as it is right now consistent in itself.

TJ-91: I think it's ok this way. We could make it more clear that our additional inputs constitute the "Domain" field in the LAMPS draft. Would that work for you, Falko?

falko-strenzke: It's OK for me, if there are complaints, we'll have to react.

falko-strenzke (approved):

draft-ietf-openpgp-pqc.md
```diff
@@ -545,34 +577,31 @@ The ML-KEM + ECDH composite public-key encryption schemes are built according to
...
+    //                   ecdhKeyShare, ecdhCipherText, ecdhPublicKey, algId)
     //
     //   Input:
     //   mlkemKeyShare   - the ML-KEM key share encoded as an octet string
     //   mlkemCipherText - the ML-KEM ciphertext encoded as an octet string
-    //   mlkemPublicKey  - The ML-KEM public key of the recipient as an octet string
+    //   mlkemPublicKey  - the ML-KEM public key of the recipient as an octet string
     //   ecdhKeyShare    - the ECDH key share encoded as an octet string
     //   ecdhCipherText  - the ECDH ciphertext encoded as an octet string
     //   ecdhPublicKey   - The ECDH public key of the recipient as an octet string
```
wussler, line 593:
```suggestion
    //   ecdhPublicKey   - the ECDH public key of the recipient as an octet string
```

Scutua (approved):

PR #166, opened 2025-02-10
Remove hash binding for ML-DSA hybrid signatures

wussler:
Proposal following discussion on the list: https://mailarchive.ietf.org/arch/msg/openpgp/8qr7qJdo6N_4eGklR6hTovtWltU/

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -752,20 +752,6 @@ Algorithm ID reference | ML-DSA    | Public key | Secret key | Signature value
 
 ## Composite Signature Schemes with ML-DSA {#ecc-mldsa}
 
-### Signature data digest {#mldsa-sig-data-digest}
-
-Signature data (i.e. the data to be signed) is digested prior to signing operations, see [[RFC9580, Section 5.2.4]](https://www.rfc-editor.org/rfc/rfc9580#section-5.2.4).
-Composite ML-DSA + EdDSA signatures MUST use the associated hash algorithm as specified in {{tab-mldsa-hash}} for the signature data digest.
-Signatures using other hash algorithms MUST be considered invalid.
-
-An implementation supporting a specific ML-DSA + EdDSA algorithm MUST also support the matching hash algorithm.
-
-{: title="Binding between ML-DSA + EdDSA and signature data digest" #tab-mldsa-hash}
-Algorithm ID reference | Hash function | Hash function ID reference
-----------------------:| ------------- | --------------------------
-30                     | SHA3-256      | 12
-31                     | SHA3-512      | 14
-
 ### Key generation procedure {#ecc-mldsa-generation}
 
 The implementation MUST generate the ML-DSA and the EdDSA component keys independently.
@@ -999,15 +985,10 @@ The input of the public keys into `multiKeyCombine` binds the KEK to the communi
 This specification makes use of the default "hedged" variants of ML-DSA and SLH-DSA, which mix fresh randomness into the respective signature-generation algorithm's internal hashing step.
 This has the advantage of an enhanced side-channel resistance of the signature operations according to  {{FIPS-204}} and {{FIPS-205}}.
 
-## Binding hashes in signatures with signature algorithms
+## Binding digest hash in signatures with SLH-DSA
 
-In order not to extend the attack surface, we bind the hash algorithm used for signature data digestion to the hash algorithm used internally by the signature algorithm.
-
-ML-DSA internally uses a SHAKE256 digest, therefore we require SHA3 in the ML-DSA + EdDSA signature packet, see {{mldsa-sig-data-digest}}.
-Note that we bind a NIST security category 2 hash function to a signature algorithm that falls into NIST security category 3.
-This does not constitute a security bottleneck: because of the unpredictable random salt that is prepended to the digested data in v6 signatures, the hardness assumption is not collision resistance but second-preimage resistance.
-
-In the case of SLH-DSA the internal hash algorithm varies based on the algorithm ID, see {{slhdsa-sig-data-digest}}.
+In order not to extend the attack surface, we bind the hash algorithm used for signature data digestion to the hash algorithm used internally by the SLH-DSA signature algorithm, see {{slhdsa-sig-data-digest}}.
+This is not done for ML-DSA to comply with the CNSA 2.0 guideline.
 
 ## Symmetric Algorithms for SEIPD Packets
 
@@ -1142,10 +1123,13 @@ TBD    | ML-KEM-1024+X448    | 56 octets X448 public key ({{tab-ecdh-cfrg-artifa
 - Fixed and improved test vectors.
 
 ## draft-ietf-openpgp-pqc-07
-- Assign code points 30 - 34 for ML-DSA and SLH-DSA algorithms.
-- Align KEM combiner with LAMPS
-- Drop CCA-conversion of X25519/X448
-- Switch to hedged variant also for SLH-DSA
+- Assigned code points 30 - 34 for ML-DSA and SLH-DSA algorithms.
+- Aligned KEM combiner with LAMPS
+- Dropped CCA-conversion of X25519/X448
+- Switched to hedged variant also for SLH-DSA
+
+## draft-ietf-openpgp-pqc-08
+- Removed hash binding for ML-DSA signatures
 
 # Contributors
 


fluppe2:
IMHO the hash bindings should not be dropped, but altered if CNSA 2.0 compliance is of utter importance.

That is, the dropped table should state SHA-384 instead of SHA3-256, and likewise SHA-512 instead of SHA3-512. On that basis compliance to CNSA 2.0 should be explicitely claimed in the "Binding hashes in signatures with signature algorithms" (keep the previous title).

Explicit is better than implicit.

wussler:
> Dropping the table means less guidance for implementers, which will result in confusion.
> 
> Explicit is better than implicit.

This is indeed true, but we don't want to disallow using SHA3 for everyone, that would also be quite nonsense IMO.
Not binding the hashes is consistent with the current OpenPGP behaviour: the choice of the hash algorithm is independent from the choice of the signature one.

I added a note in the security considerations about this, but maybe we can make it more explicit.

Other option, is to keep the binding but allow both SHA2 and SHA3:
```
{: title="Binding between ML-DSA + EdDSA and signature data digest" #tab-mldsa-hash}
Algorithm ID reference | Hash function | Hash function ID reference
----------------------:| ------------- | --------------------------
30                     | SHA2-256 SHA3-256      | 8, 12
31                     | SHA2-512 SHA3-512      | 10, 14
```
But I personally don't see the point that much

draft-ietf-openpgp-pqc.md
```diff
@@ -999,15 +985,10 @@ The input of the public keys into `multiKeyCombine` binds the KEK to the communi
...
-In order not to extend the attack surface, we bind the hash algorithm used for signature data digestion to the hash algorithm used internally by the signature algorithm.
+## Binding digest hash in signatures with SLH-DSA
 
-ML-DSA internally uses a SHAKE256 digest, therefore we require SHA3 in the ML-DSA + EdDSA signature packet, see {{mldsa-sig-data-digest}}.
-Note that we bind a NIST security category 2 hash function to a signature algorithm that falls into NIST security category 3.
-This does not constitute a security bottleneck: because of the unpredictable random salt that is prepended to the digested data in v6 signatures, the hardness assumption is not collision resistance but second-preimage resistance.
-
-In the case of SLH-DSA the internal hash algorithm varies based on the algorithm ID, see {{slhdsa-sig-data-digest}}.
+In order not to extend the attack surface, we bind the hash algorithm used for signature data digestion to the hash algorithm used internally by the SLH-DSA signature algorithm, see {{slhdsa-sig-data-digest}}.
+This is not done for ML-DSA to comply with the CNSA 2.0 guideline.
```
TJ-91, line 991: I would not argue for ML-DSA in the SLH-DSA section. I would either move this to its own section or rename the title s.t. it does not exclude ML-DSA.

fluppe2:
As I understand RFC9580 [5.2.4. ](https://www.rfc-editor.org/rfc/rfc9580#section-5.2.4)[Computing Signatures](https://www.rfc-editor.org/rfc/rfc9580#name-computing-signatures), the hash algorithm ID is part of the trailer.

Therefore I would rather like to explicitly list the allowed options as you suggested. Keep in mind that CNSA 2.0 states "Use SHA-384 or SHA-512 for all classification levels.". I think that means SHA2-384 or SHA2-512 instead of SHA2-256 in the table.

draft-ietf-openpgp-pqc.md
wussler:
```suggestion
```

draft-ietf-openpgp-pqc.md
```diff
@@ -752,20 +752,6 @@ Algorithm ID reference | ML-DSA    | Public key | Secret key | Signature value
 
 ## Composite Signature Schemes with ML-DSA {#ecc-mldsa}
```
wussler, line 754:
```suggestion

### Signature data digest {#mldsa-sig-data-digest}
In order not to extend the attack surface, composite ML-DSA + EdDSA signatures SHOULD use SHA3 for the signature data digest because ML-DSA internally uses a SHAKE256 digest.
Implementations MAY allow use of SHA2 for CNSA 2.0 compliance.
```

TJ-91:
Since we deviate from the default OpenPGP behaviour, perhaps we should also add a statement that flexibility in the digest algorithms may hurt security when matching "too weak" hash functions. While it historically may have made sense to have flexibility, we do not see a reason to have it with the PQC algorithms.

For SLH-DSA we can even enhance the argument why using the matching SHA3 function increases security. In contast to most other signature schemes, its security relies entirely upon the security properties of a hash function and no additional hardness assumptions about other mathematical problems are required. Thus, the attack surface is (relatively speaking) much more affected when using a different hash function.

The reason why I think it makes sense to add these things is that previously we had a clear story ("match the internally used hash algorithms"), but now we are more lenient with ML-DSA and it's less consistent. However, if you think the current state is sufficient, I'm also fine with it. Also, it can be part of the security considerations, I think.

falko-strenzke:
> Since we deviate from the default OpenPGP behaviour, perhaps we should also add a statement that flexibility in the digest algorithms may hurt security when matching "too weak" hash functions. While it historically may have made sense to have flexibility, we do not see a reason to have it with the PQC algorithms.
> 
> For SLH-DSA we can even enhance the argument why using the matching SHA3 function increases security. In contast to most other signature schemes, its security relies entirely upon the security properties of a hash function and no additional hardness assumptions about other mathematical problems are required. Thus, the attack surface is (relatively speaking) much more affected when using a different hash function.
> 
> The reason why I think it makes sense to add these things is that previously we had a clear story ("match the internally used hash algorithms"), but now we are more lenient with ML-DSA and it's less consistent. However, if you think the current state is sufficient, I'm also fine with it. Also, it can be part of the security considerations, I think.

What I see as a concrete formal loss is that we loose the resistance against digest substitution attacks. But I think this is not an issue that has any practical relevance.

draft-ietf-openpgp-pqc.md
```diff
@@ -752,6 +752,10 @@ Algorithm ID reference | ML-DSA    | Public key | Secret key | Signature value
...
 
+### Signature data digest {#mldsa-sig-data-digest}
+In order not to extend the attack surface, composite ML-DSA + EdDSA signatures SHOULD use SHA3 for the signature data digest because ML-DSA internally uses a SHAKE256 digest.
```
fluppe2, line 756: SHA2-384 and SHA2-512 are internationally approved hash functions and therefore a viable choice to compute a message digest. We should not claim or hint at, that the use of those SHA2 instances for the signature data digest extends the attack surface.

TJ-91: It's not about security concerns with SHA2 specifically but about extending the attack surface due to adding an additional hash algorithm from a different algorithm family to the signature operation (if the digest algorithm doesn't match the internally used algorithm). 

Maybe it needs better wording?

draft-ietf-openpgp-pqc.md
```diff
@@ -755,16 +755,17 @@ Algorithm ID reference | ML-DSA    | Public key | Secret key | Signature value
 ### Signature data digest {#mldsa-sig-data-digest}
 
 Signature data (i.e. the data to be signed) is digested prior to signing operations, see [[RFC9580, Section 5.2.4]](https://www.rfc-editor.org/rfc/rfc9580#section-5.2.4).
-Composite ML-DSA + EdDSA signatures MUST use the associated hash algorithm as specified in {{tab-mldsa-hash}} for the signature data digest.
-Signatures using other hash algorithms MUST be considered invalid.
+Composite ML-DSA + EdDSA signatures MUST use an associated hash algorithm as specified in {{tab-mldsa-hash}} for the signature data digest. They SHOULD use SHA3 to match the ML-DSA internal use of SHAKE256. They MAY use SHA2 for CNSA 2.0 compliance. Signatures using other combinations or hash algorithms MUST be considered invalid.
```
falko-strenzke, line 758: I would not write anything about CNSA. It we allow SHA-2, we allow it, I don't think it makes sense to put that under further requirements (otherwise it ends up in a weird mixture of conditional MAY and MUST requirements).

 I would write:
> They MAY use SHA2. This can for instance be necessary to satisfy specific regulatory requirements.

draft-ietf-openpgp-pqc.md
```diff
@@ -1142,10 +1137,13 @@ TBD    | ML-KEM-1024+X448    | 56 octets X448 public key ({{tab-ecdh-cfrg-artifa
...
-- Align KEM combiner with LAMPS
-- Drop CCA-conversion of X25519/X448
-- Switch to hedged variant also for SLH-DSA
+- Assigned code points 30 - 34 for ML-DSA and SLH-DSA algorithms.
+- Aligned KEM combiner with LAMPS
+- Dropped CCA-conversion of X25519/X448
+- Switched to hedged variant also for SLH-DSA
+
+## draft-ietf-openpgp-pqc-08
+- Removed hash binding for ML-DSA signatures
```
wussler, line 1146:
```suggestion
- Changed hash binding for ML-DSA signatures
```

draft-ietf-openpgp-pqc.md
wussler, line 758:
```suggestion
Composite ML-DSA + EdDSA signatures MUST use an associated hash algorithm as specified in {{tab-mldsa-hash}} for the signature data digest. They SHOULD use SHA3 to match the ML-DSA internal use of SHAKE256. They MAY use SHA2, this can for instance be necessary to satisfy specific regulatory requirements. Signatures using other combinations or hash algorithms MUST be considered invalid.
```

fluppe2:
This looks good to me. Did some minor cleanups also. Please comment and/or approve.

Do we need to specify now the hash function that has been used for the signature data digest in test vector A.4?

fluppe2 (approved):

wussler (commented):
I think the refactor messed it up the display a little, but looking at the diff it looks good to me. Can't approve because am author

draft-ietf-openpgp-pqc.md
```diff
@@ -999,15 +951,10 @@ The input of the public keys into `multiKeyCombine` binds the KEK to the communi
...
 
-ML-DSA internally uses a SHAKE256 digest, therefore we require SHA3 in the ML-DSA + EdDSA signature packet, see {{mldsa-sig-data-digest}}.
-Note that we bind a NIST security category 2 hash function to a signature algorithm that falls into NIST security category 3.
-This does not constitute a security bottleneck: because of the unpredictable random salt that is prepended to the digested data in v6 signatures, the hardness assumption is not collision resistance but second-preimage resistance.
-
-In the case of SLH-DSA the internal hash algorithm varies based on the algorithm ID, see {{slhdsa-sig-data-digest}}.
+Signature data (i.e. the data to be signed) is digested prior to signing operations, see [[RFC9580, Section 5.2.4]](https://www.rfc-editor.org/rfc/rfc9580#section-5.2.4).
+Binding a NIST security category 2 hash function (e.g. SHA3-256) to a signature algorithm that falls into NIST security category 3 or 5 does not constitute a security bottleneck: because of the unpredictable random salt that is prepended to the digested data in v6 signatures, the hardness assumption is not collision resistance but second-preimage resistance.
```
TJ-91, line 957: 1. I would not use the term "binding" any more.
2. Since we do not change anything in the OpenPGP protocol we do not need to reiterate that signature data is hashed before signing, At least the reference to RFC 9580 is overkill IMO.
3. The section seems to be about NIST security categories and how the salt helps to achieve higher security levels with "weaker" hash functions. There is no reference or introduction to the NIST security categories. I think that would be needed.

Suggestion: Rework the section and title, make it about what NIST security categories are achieved.

Alternative suggestion: Just remove it entirely. I think the benefit of the salt regarding collision / second preimage resistance is already part of RFC 9580. I'm not sure how relevant the NIST security categories are for this document.

Third alternative: Just add a one-sentence statement about the salt and second-preimage resistance somewhere.

fluppe2: I just boiled down already existing text. But I see that it does not add that much anymore, now that we have removed the bindings. I would go with the alternative suggestion and remove the section completely, so that we simply add new signature algorithms build upon RFC9580.

fluppe2 (approved):

TJ-91 (approved):

dignifiedquire:
I just want to note that from an implementors and users perspective, switching to full hash agility is very disappointing..

falko-strenzke:
> I just want to note that from an implementors and users perspective, switching to full hash agility is very disappointing..

Can you say what is your exact – security-related I assume – concern?

falko-strenzke:
It is a principle weakening, that is correct. However, the random salt that precedes the signed message in v6 signatures mitigates that to a large extent: Hash collision attacks are prevented by this. With the salt, even the use of MD5 would be secure as of today. What remains is only the threat of 2nd-preimage attacks. Such attacks are not known for any practically relevant hash functions. 

With the second point about interoperability that is probably true. But in any case with the change the situation for post-quantum signatures is as it is for traditional signatures... I think signers and verifiers simply have to set reasonable rules for the acceptable hash functions anyway.

PR #167, opened 2025-02-11
Remove normative guidance about key selection

### LINKED ISSUE #165: §8.1 contains normative guidance about key selection
In the [interim meeting](https://datatracker.ietf.org/meeting/interim-2025-openpgp-01/session/openpgp) today, some discussion around PQC's §8.1's guidance on key selection.

I think two related concerns were raised:

- The section talks about "preferring" PQ/T  keys in the abstract over non-PQ/T keys, but then shifts into talking about encryption; does the same preference apply to signing/verifying?  i don't think it even makes sense to apply to decryption, but i guess that's worth asking as well.
- The section seems to take a normative stance on selecting a single subkey for encryption, but there may be scenarios where that's not advisable for usability, even if it means giving up on the PQ resistance for the message in question.

Perhaps the right way forward is to remove the normative guidance about key selection entirely, and just leave the note about how the post-quantum resistance only applies *if* a given message has not also been encrypted to a non-PQ key.

TJ-91:
closes #165 

I split the migration considerations into the signing and encrypting case and remove the normative statements about which keys to prefer.

Further, I added some context in the security considerations, especially the "store now, decrypt later" attack scenario, since I feel this is the most important aspect in the current situation.

I am happy for suggestions and improvements

---

> An implementation MAY sign with both a PQ(/T) and an ECC key [...]

Here I changed "ECC" to "traditional" or did we have any specific reason why we write "ECC" instead?

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -936,7 +936,7 @@ An OpenPGP certificate is composed of a certification-capable primary key and on
 Two migration strategies are recommended:
 
 1. Generate two independent certificates, one for PQ(/T)-capable implementations, and one for legacy implementations.
-   Implementations not understanding PQ(/T) certificates can use the legacy certificate, while PQ(/T)-capable implementations will prefer the newer certificate.
+   Implementations not understanding PQ(/T) certificates can use the legacy certificate, while PQ(/T)-capable implementations can also use the newer certificate.
    This allows having a traditional certificate for compatibility and a v6 PQ(/T) certificate, at a greater complexity in key distribution.
 
 2. Attach PQ(/T) encryption or signature subkeys to an existing traditional v6 OpenPGP certificate.
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -945,17 +945,6 @@ Two migration strategies are recommended:
 
 # Security Considerations
 
-## Parallel Encryption to PQ/(T) and Traditional Keys
-
-This document does not specify how implementations should handle encrypting to PQ(/T) keys and traditional keys at the same time, and it is out of scope to define which encryption subkeys should be selected.
-However, implementers should be aware of the implications.
-
-If at least one PKESK packet for an encrypted message uses a traditional public-key algorithm, the message can be decrypted by a cryptographically relevant quantum computer (CRQC).
-While the existence of a CRQC is not yet known to the public at the time of writing this document, they may become a reality in the following years.
-This also affects the confidentiality of messages sent today due to the possibility of storing traditionally-encrypted messages and decrypting them later when a CRQC is available.
-This attack scenario is commonly referred to as "store now, decrypt later" or "harvest now, decrypt later".
-While ideally all implementations should switch to exclusively encrypting with PQ/T algorithms from a security standpoint, this breaks interoperability with all currently deployed traditional certificates and implementations that do not yet implement the PQ/T algorithms.
-
 ## Security Aspects of Composite Signatures
 
 When multiple signatures are applied to a message, the question of the protocol's resistance against signature stripping attacks naturally arises.
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -917,7 +917,7 @@ The post-quantum KEM algorithms defined in {{kem-alg-specs}} and the signature a
 During the transition period, the post-quantum algorithms will not be supported by all clients.
 Therefore various migration considerations must be taken into account, in particular backwards compatibility to existing implementations that have not yet been updated to support the post-quantum algorithms.
 
-## Encrypting to Traditional and PQ/T Keys
+## Encrypting to Traditional and PQ(/T) Keys
 
 As noted in {{composite-kem}}, the confidentiality of a message is not post-quantum secure when using multiple PKESKs if at least one does not use PQ/T encryption schemes.
 An implementation SHOULD NOT abort the encryption process when encrypting a message to both PQ/T and traditional keys to allow for a smooth transition to post-quantum cryptography.
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -953,7 +953,7 @@ However, implementers should be aware of the implications.
 If at least one PKESK packet for an encrypted message uses a traditional public-key algorithm, the message can be decrypted by a cryptographically relevant quantum computer (CRQC).
 While the existence of a CRQC is not yet known to the public at the time of writing this document, they may become a reality in the following years.
 This also affects the confidentiality of messages sent today due to the possibility of storing traditionally-encrypted messages and decrypting them later when a CRQC is available.
-This strategy is commonly referred to as "store now, decrypt later" or "harvest now, decrypt later".
+This attack scenario is commonly referred to as "store now, decrypt later" or "harvest now, decrypt later".
 While ideally all implementations should switch to exclusively encrypting with PQ/T algorithms from a security standpoint, this breaks interoperability with all currently deployed traditional certificates and implementations that do not yet implement the PQ/T algorithms.
 
 ## Security Aspects of Composite Signatures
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -949,6 +945,17 @@ Two migration strategies are recommended:
 
 # Security Considerations
 
+## Parallel Encryption to PQ/(T) and Traditional Keys
+
+This document does not specify how implementations should handle encrypting to PQ(/T) keys and traditional keys at the same time, and it is out of scope to define which encryption subkeys should be selected.
+However, implementers should be aware of the implications.
+
+If at least one PKESK packet for an encrypted message uses a traditional public-key algorithm, the message can be decrypted by a cryptographically relevant quantum computer (CRQC).
+While the existence of a CRQC is not yet known to the public at the time of writing this document, they may become a reality in the following years.
```
falko-strenzke, line 954:
```suggestion
While the existence of a CRQC is not yet known to the public at the time of writing this document, they may become a reality in the not too distant future.
```
"In the following years" is too much of a prediction in my view.

draft-ietf-openpgp-pqc.md
```diff
@@ -949,6 +945,17 @@ Two migration strategies are recommended:
...
+This also affects the confidentiality of messages sent today due to the possibility of storing traditionally-encrypted messages and decrypting them later when a CRQC is available.
+This strategy is commonly referred to as "store now, decrypt later" or "harvest now, decrypt later".
```
falko-strenzke, line 956:
```suggestion
This attack scenario is commonly referred to as "store now, decrypt later" or "harvest now, decrypt later".
```

draft-ietf-openpgp-pqc.md
TJ-91: For me "following years" is less of a prediction than "not too distant future". Both are fairly subjective, so how about writing just "future" instead of "not too distrant future"?

draft-ietf-openpgp-pqc.md
```diff
@@ -918,17 +917,14 @@ The post-quantum KEM algorithms defined in {{kem-alg-specs}} and the signature a
...
-## Key preference {#pq-key-preference}
+## Encrypting to Traditional and PQ/T Keys
 
-Implementations SHOULD prefer PQ(/T) keys when multiple options are available.
-When encrypting to a certificate that has both a valid PQ/T and a valid traditional encryption subkey, an implementation SHOULD use the PQ/T subkey only.
-Furthermore, if an application has any means to determine that encrypting to a PQ/T certificate and a traditional certificate is redundant, it should omit encrypting to the traditional certificate.
-
-As specified in {{composite-kem}}, the confidentiality of a message is not post-quantum secure when using multiple PKESKs if at least one does not use PQ/T encryption schemes.
+As noted in {{composite-kem}}, the confidentiality of a message is not post-quantum secure when using multiple PKESKs if at least one does not use PQ/T encryption schemes.
 An implementation SHOULD NOT abort the encryption process when encrypting a message to both PQ/T and traditional keys to allow for a smooth transition to post-quantum cryptography.
```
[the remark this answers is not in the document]
TJ-91: I am not 100% decided on this what do the others think?

draft-ietf-openpgp-pqc.md
```diff
@@ -918,17 +917,14 @@ The post-quantum KEM algorithms defined in {{kem-alg-specs}} and the signature a
 During the transition period, the post-quantum algorithms will not be supported by all clients.
 Therefore various migration considerations must be taken into account, in particular backwards compatibility to existing implementations that have not yet been updated to support the post-quantum algorithms.
 
-## Key preference {#pq-key-preference}
+## Encrypting to Traditional and PQ/T Keys
```
fluppe2, line 920:
```suggestion
## Encrypting to Traditional and PQ(/T) Keys
```

draft-ietf-openpgp-pqc.md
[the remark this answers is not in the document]
fluppe2: I think we should remove the sentence.

draft-ietf-openpgp-pqc.md
```diff
@@ -949,6 +945,17 @@ Two migration strategies are recommended:
 
 # Security Considerations
 
+## Parallel Encryption to PQ/(T) and Traditional Keys
```
fluppe2, line 948:
```suggestion
## Parallel Encryption to PQ(/T) and Traditional Keys
```

draft-ietf-openpgp-pqc.md
TJ-91: The convention in the document seems to be to use `PQ/T` for encryption and `PQ(/T)` for signatures, since all encryption algorithms are `PQ/T` but for signatures we have `PQ` and `PQ/T` algorithms. I'm not sure if that's really necessary and I would not be against changing everything to PQ(/T) (in a separete PR)

draft-ietf-openpgp-pqc.md
```diff
@@ -949,6 +945,17 @@ Two migration strategies are recommended:
...
+
+This document does not specify how implementations should handle encrypting to PQ(/T) keys and traditional keys at the same time, and it is out of scope to define which encryption subkeys should be selected.
+However, implementers should be aware of the implications.
+
+If at least one PKESK packet for an encrypted message uses a traditional public-key algorithm, the message can be decrypted by a cryptographically relevant quantum computer (CRQC).
+While the existence of a CRQC is not yet known to the public at the time of writing this document, they may become a reality in the following years.
+This also affects the confidentiality of messages sent today due to the possibility of storing traditionally-encrypted messages and decrypting them later when a CRQC is available.
```
falko-strenzke, line 955: I wouldn't refer to "today" in an RFC. 
```suggestion
This also affects the confidentiality of messages sent in the pre-quantum era due to the possibility of storing traditionally-encrypted messages and decrypting them later when a CRQC is available.
```

TJ-91:
Thanks for the input everybody. Do we all agree with the current state or are there still things we need to change?

draft-ietf-openpgp-pqc.md
```diff
@@ -918,17 +917,14 @@ The post-quantum KEM algorithms defined in {{kem-alg-specs}} and the signature a
...
+As noted in {{composite-kem}}, the confidentiality of a message is not post-quantum secure when using multiple PKESKs if at least one does not use PQ(/T) encryption schemes.
+An implementation should not abort the encryption process when encrypting a message to both PQ/T and traditional keys to allow for a smooth transition to post-quantum cryptography.
 
-As specified in {{composite-kem}}, the confidentiality of a message is not post-quantum secure when using multiple PKESKs if at least one does not use PQ/T encryption schemes.
-An implementation SHOULD NOT abort the encryption process when encrypting a message to both PQ/T and traditional keys to allow for a smooth transition to post-quantum cryptography.
+## Signing with Traditional and PQ(/T) Keys
 
-
-An implementation MAY sign with both a PQ(/T) and an ECC key using multiple signatures over the same data as described in {{multiple-signatures}}.
+An implementation MAY sign with both a PQ(/T) and a traditional key using multiple signatures over the same data as described in {{multiple-signatures}}.
```
fluppe2, line 927:
```suggestion
An implementation may sign with both a PQ(/T) and a traditional key using multiple signatures over the same data as described in {{multiple-signatures}}.
```
I would also put this in lower case to avoid normative language in the considerations section.

fluppe2 (changes requested):
We have more normative language in the "Key generation strategies" section. That is,

"It is RECOMMENDED to generate fresh secrets when generating PQ(/T) keys."

I would propose to put that in lower case also.

TJ-91:
I think we can have normative statements in that section or can we really not? Since it's not touching anything related to trust model or choosing keys, I would keep it like it is. @wussler what do you think?

TJ-91:
in [b0bfefe](https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/167/commits/b0bfefe517df67f8aeaffa021f822a5608906e12) I reworded s.t. no preference for PQC keys is implied

fluppe2 (approved):

fluppe2:
Fine for me.

wussler:
I think this PR requires a line in the changelog. Rest LGTM

wussler (approved):

PR #168, opened 2025-02-13
Assign ML-KEM code points and remove some notes to the reader

TJ-91:


fluppe2:
We have another "Note to the reader" in section "3.3. Multiple Signatures". Is that to be removed as well?

fluppe2 (changes requested):
Looks good to me.

We have another "Note to the reader" in section "3.3. Multiple Signatures". Is that to be removed as well before merging?

TJ-91:
I reworded it is that better?

fluppe2 (approved):

fluppe2:
Yes, for me it's fine.

wussler (approved):

PR #179, opened 2025-04-10
Update test vectors

### LINKED ISSUE #170: Update Test Vectors
After #177 is resolved

wussler:
- Updates existing test vectors
- Adds v4 test vectors
- Adds missing SLH-DSA test vectors

Closes #170


TJ-91:
Thanks for taking care of this! I can import all keys and decrypt all messages correctly.

Some remarks:

* For v6 eddsa we should now switch to v2 SEIPD, I think, since we have v1 SEIPD with v4 eddsa already.
* For SLH-DSA 128f and 256s we only have the sk as test data. Maybe we agreed on doing it like this, I don't remember, butI want to make sure it's not an accident. Having only the sk is probably enough to confirm a correct implementation given we already have full test vectors for 128s. I would still prefer to also include the pk and a message. Of course this roughly doubles the document size.

wussler:
> For v6 eddsa we should now switch to v2 SEIPD, I think, since we have v1 SEIPD with v4 eddsa already.

I had considered this, but there is already a v2 SEIPD for the same algorithm, and it is still a valid case to have V6+SEIPDv1

>For SLH-DSA 128f and 256s we only have the sk as test data. Maybe we agreed on doing it like this, I don't remember, butI want to make sure it's not an accident. Having only the sk is probably enough to confirm a correct implementation given we already have full test vectors for 128s. I would still prefer to also include the pk and a message. Of course this roughly doubles the document size.

I am not worried about the size, but this vectors are already complete IMO. 
The keys contain already signatures for these algorithms, and the encrypted messages would be a repetition of what we had already in the previous vectors. The SKs contain also the PK data.

If you (@TJ-91 and @dkg) believe it adds value to generate also PKs and messages, happy to do so.

dkg:
I think for signing keys it is useful to have an explicit form of:

- secret key material (TSKs)
- public key material (TPKs)
- detached signature over a simple message

This makes it so that any implementer can immediately test, directly from the draft, using some invocations like:

- `sopv verify msg.sig sample.cert < msg`
- `sop extract-cert < sample.key > sample2.cert`
- `cmp <(sop dearmor < sample.cert) <(sop dearmor < sample2.cert)`

hko-s:
I just updated to these test vectors in an rPGP PQC branch (see https://github.com/rpgp/rpgp/pull/533/).

Based on this, I'd say these new vectors look good.
Specifically, the rPGP tests check:

- fingerprints of all primaries and subkeys as shown in the updated draft text
- validity of binding self-signatures (including for the two new SLH-DSA TSKs)
- message decryption and data signature verification

However, the rPGP tests do not verify the intermediate hex values (key shares, key combiner output, session keys)!
So I can't comment on the correctness of those values in the text.

wussler:
Thank you for your feedback @dkg and @hko-s!

I will
- add the public key material (TPKs) for the SLH-DSA keys (where missing)
- add a detached signature over a simple message for all PQ signing keys

To test the intermediate values, we'll rely on @TJ-91's testing

wussler:
Added the missing test vectors. Please @TJ-91 and @falko-strenzke have a look!

hko-s:
> Added the missing test vectors. Please @TJ-91 and @falko-strenzke have a look!

I tested these vectors in https://github.com/rpgp/rpgp/pull/534 and found no issues

TJ-91 (approved):
Looks good to me! I also checked the intermediate KEM combiner values and fingerprints etc.

Also, thanks for checking with rpgp!

PR #182, opened 2025-05-02
Attempt at clarifying certificate structure

hko-s:
A primary key is implicitly capable of issuing self-signatures for certificate formation and life cycle management (this might be best omitted in this text)

It may be marked "certification"-capable, which allows it to issue valid third-party certifications (however, primary keys may exist without this capability)

The primary can in principle fulfill all capabilities (however, except for the case of RSA, it can't in practice be used for encryption, when its asymmetric mechanism only supports signing operations). Still, even with modern keys, the primary may be used for data signatures and authentication (and at least in some special-purposes scenarios, this may be a reasonable approach)

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -842,7 +842,7 @@ Signing only with PQ(/T) key material is not backwards compatible.
 It is RECOMMENDED to generate fresh secrets when generating PQ(/T) keys.
 Note that reusing key material from existing ECC keys in PQ(/T) keys does not provide backwards compatibility.
 
-An OpenPGP certificate is composed of a certification-capable primary key and one or more subkeys for signature, encryption, and authentication.
+An OpenPGP certificate is composed of a primary key which is often marked as capable of issuing third-party certifications, and possibly other additional capabilities, and one or more subkeys for data signatures, encryption, and authentication.
 Two migration strategies are recommended:
 
 1. Generate two independent certificates, one for PQ(/T)-capable implementations, and one for legacy implementations.


draft-ietf-openpgp-pqc.md
```diff
@@ -842,7 +842,7 @@ Signing only with PQ(/T) key material is not backwards compatible.
 It is RECOMMENDED to generate fresh secrets when generating PQ(/T) keys.
 Note that reusing key material from existing ECC keys in PQ(/T) keys does not provide backwards compatibility.
 
-An OpenPGP certificate is composed of a certification-capable primary key and one or more subkeys for signature, encryption, and authentication.
+An OpenPGP certificate is composed of a primary key which is often marked capable of issuing third-party certifications, and possibly other additional capabilities, and one or more subkeys for data signatures, encryption, and authentication.
```
falko-strenzke, line 845:
```suggestion
An OpenPGP certificate is composed of a primary key which is often marked as capable of issuing third-party certifications, and possibly other additional capabilities, and one or more subkeys for data signatures, encryption, and authentication.
```

hko-s: I changed this, thanks!

TJ-91:
I think "one or more subkeys" is technically wrong since subkeys are entirely optional.

Aside from that I'm not sure whether there really is a benefit in reiterating the certificate structure here and going into this level of detail about the key capabilities. We might just drop the sentence to keep it simple.

hko-s:
> I think "one or more subkeys" is technically wrong since subkeys are entirely optional.

Agreed, that should actually read "zero or more".

> Aside from that I'm not sure whether there really is a benefit in reiterating the certificate structure here and going into this level of detail about the key capabilities. We might just drop the sentence to keep it simple.

I have no strong feelings against dropping the sentence. I was mainly concerned by having an explanation that could confuse readers. 

However, it does seem helpful to remind readers that an OpenPGP certificate is a complex data structure that often contains multiple keys, whose mutual relationships have some implications for migration. But keeping this as brief as possible sounds good to me.

falko-strenzke:
The relevant sentence in the draft was removed by #185 and thus we closed this PR.

The authors will follow up with an explanation on the mailing list.  

@hko-s Feel free to open an issue if you think something was not addressed.

hko-s:
Thanks, works for me! :+1: 

While I'm sad that the draft has been losing a lot of guidance since -07, I'd like to see it released sooner rather than later. If that means losing some guidance in the text, I'm a bit sad, but still think it's the least bad trade-off.

PR #184, opened 2025-05-05
Update test vectors

wussler:
As discussed on the list, this MR updates the test vectors:
- Removing the ECDH subkeys
- Adding an SEIPDv2 to the V4 test vectors to cover both cases
- Replacing the SEIPDv1 with an SEIPDv2 for the V6 ECC test vector

Furthermore, we update the key creation time with something more reasonable


TJ-91:
I took the liberty to add trailing newlines to the files since RNP is a bit picky.
I checked and everything verifies. The intermediate values match and I could not see anything apart from the following two things:

For `Sample v4 Ed25519 with ML-KEM-768+X25519 Data` it says there is a v4 direct key self-signature for the sk and pk. There is no direct key signature, though.

----

> Sample SLH-DSA-128s with ML-KEM-768+X25519 Data

> Sample SLH-DSA-128f with ML-KEM-768+X25519 Data

> Sample SLH-DSA-256s with ML-KEM-1024+X448 Data


The algorithm names include "SHAKE":
* SLH-DSA-SHAKE-128s
* SLH-DSA-SHAKE-128f
* SLH-DSA-SHAKE-256s

Therefore I would include it here too.

falko-strenzke:
> As discussed on the list, this MR updates the test vectors:
> 
>     * Removing the ECDH subkeys
>

Why are removing the ECDH subkeys? As far as I can see, their presence was not a concern for Justus. In my view, the ECDH subkeys allow an important test: namely whether a legacy implementation can work with the certificates containing PQC encryption subkeys.

twiss:
My main concern with the current draft and/or test vectors as I wrote [here](https://mailarchive.ietf.org/arch/msg/openpgp/ipOOKb5TKii_xtNxe3gifF50d2w/) is that the expected behavior for _new_ implementations (that do support PQC) is not specified anywhere. Are they supposed to encrypt to the PQC subkey, the ECC subkey, or both? It's obvious that only the first option gives post-quantum security, and some parts of the draft seem to imply that it should be possible to achieve that, but it's not spelled out (since draft-08) and the implementations are not consistent.
I would like to be able to encode, in the interop test suite, what the expected behavior when using the test vectors is, so that we can aim at a more consistent behavior.
My preferred solution is (now) to add back some guidance (i.e. option 2 in the email linked above) but updating the test vectors and changing the text of the draft (option 1) is also possible.

falko-strenzke:
> My main concern with the current draft and/or test vectors as I wrote [here](https://mailarchive.ietf.org/arch/msg/openpgp/ipOOKb5TKii_xtNxe3gifF50d2w/) is that the expected behavior for _new_ implementations (that do support PQC) is not specified anywhere. Are they supposed to encrypt to the PQC subkey, the ECC subkey, or both?

We had reached consenus in the WG that this draft should not specify the encryption subkey selection, as this was intended to be covered by seperate specifications. This is why all (concrete) guidance on subkey selection was removed (maybe even still more has to be removed, that's possible) in version 08 (admitedly, this change is missing from the changelog in the draft) . I don't think I have to collect the references to the disussion here. Otherwise I can do so, but I assume we all remember this decision well. As a natural result, the draft intentionally does *not* specify the subkey selection.

If you hereby want to declare that you now disagree with this choice, the logical consequence is that we must seek a new consensus. But we can't simply move the draft in the direction that you request now, as other WG members obviously have stated a different opinion before.

> It's obvious that only the first option gives post-quantum security, and some parts of the draft seem to imply that it should be possible to achieve that, but it's not spelled out (since draft-08) and the implementations are not consistent.

As I explained above, based on the (at least previously existing) consensus, implementers can't expect concrete guidance from the PQC draft in this respect.

> I would like to be able to encode, in the interop test suite, what the expected behavior when using the test vectors is, so that we can aim at a more consistent behavior. My preferred solution is (now) to add back some guidance (i.e. option 2 in the email linked above) but updating the test vectors and changing the text of the draft (option 1) is also possible.

As I said above, this requires to find a new consensus in the WG. If you uphold this aim on the list, I assume that the chairs will be dealing with this question at the end of the WGLC.

twiss:
The main point of my email was to point out an inconsistency between the draft and what we had decided. There are two obvious options: change the draft or change the decision. I won't insist on one of the two; I expressed my preference but that's all it is. If you insist on the former I won't object formally.

> maybe even still more has to be removed, that's possible

I'm not sure why you say "maybe", I clearly quoted the inconsistent passages in my email, no?

twiss:
Back to the main topic of the test vectors: if you go the route of removing all key selection guidance, then IMO the test vectors with multiple subkeys are a bit meaningless because it's not clear what to test there or how a new implementation should behave when using them.

You said:

> In my view, the ECDH subkeys allow an important test: namely whether a legacy implementation can work with the certificates containing PQC encryption subkeys.

However, IMO this is not the primary purpose of test vectors in the draft, those should serve new implementations that are trying to implement the draft.

To test how legacy implementations behave (when using mixed test vectors, for example) we can add additional test vectors to the interoperability test suite, beyond those in the draft (as we already do today).

falko-strenzke:
> Back to the main topic of the test vectors: if you go the route of removing all (references to) key selection guidance, then IMO the test vectors with multiple subkeys are a bit meaningless because it's not clear what to test there or how a new implementation should behave when using them.
> 
> You said:
> 
> > In my view, the ECDH subkeys allow an important test: namely whether a legacy implementation can work with the certificates containing PQC encryption subkeys.
> 
> However, IMO this is not the primary purpose of test vectors in the draft, those should serve new implementations that are trying to implement the draft.
> 
> To test how legacy implementations behave (when using mixed test vectors, for example) we can add additional test vectors to the interoperability test suite, beyond those in the draft (as we already do today).

It's OK for me if the ECDH encryption subkeys are removed.

falko-strenzke:
> The main point of my email was to point out an inconsistency between the draft and what we had decided. There are two obvious solutions: change the draft or change the decision. I won't insist on one of the two; I expressed my preference but that's all it is. If you insist on the former I won't object formally.
> 
> > maybe even still more has to be removed, that's possible
> 
> I'm not sure why you say "maybe", I clearly quoted the inconsistent passages in my email, no?

Where exactly do you see the inconsistency? If it is about the remaining statements about subkey selection, then I agree. In my view it was decided not to have any such statements in the draft.

wussler:
This PR now aligns with the simplified guidance in #185

falko-strenzke (approved):

fluppe2 (approved):
LGTM (without explicit re-calc of testvecs)

PR #188, opened 2025-05-22
Specify minimum requirements for signature digest sizes

TJ-91:
restrict hash algorithms for PQC signature algorithms similar to what RFC 9580 specifies, based on Daniel Huigen's suggestion in https://mailarchive.ietf.org/arch/msg/openpgp/r0C-dXz35BKx5rDsznt4rGcS6pM/

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -894,7 +894,7 @@ This has the advantage of an enhanced side-channel resistance of the signature o
 ## Minimum digest size for PQ(/T)-signatures
 
 This specification requires that all PQ(/T) signatures defined in this document are made in combination with a hash algorithm of at least 256 bits digest size.
-The relevant security property is preimage resistance, since all signature algorithms defined in this document require signature packets of version 6 or higher, which include a random salt value.
+The relevant security property is preimage resistance, since all signature algorithms defined in this document require signature packets of version 6 or higher, which currently include a random salt prefix.
 Therefore, a hash algorithm with 256 bits digest size is sufficient to match the targeted security level of all PQ(/T) algorithms defined in this document.
 Note that the minimum hash algorithm digest size requirement is different for signature schemes that are defined in [RFC9580], where signature algorithms are also allowed to be used with non-salted v4 signatures.
 In such cases, the relevant security property is collision resistance.
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -795,7 +795,7 @@ The algorithm-specific part of a signature packet for an SLH-DSA algorithm code 
 
  - A fixed-length octet string of the SLH-DSA signature value, whose length depends on the algorithm ID in the format specified in {{slhdsa-artifact-lengths}}.
 
-An SLH-DSA signature MUST use a hash algorithm with a digest size of at least 256 bits.
+An SLH-DSA signature MUST use a hash algorithm with a digest size of at least 256 bits for the computation of the message digest.
 A verifying implementation MUST reject any SLH-DSA signature that uses a hash algorithm with a smaller digest size.
 
 ### Key Material Packets
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -735,7 +735,7 @@ The algorithm-specific v6 signature parameters for ML-DSA + EdDSA signatures con
 
  - A fixed-length octet string of the ML-DSA signature value, whose length depends on the algorithm ID as specified in {{tab-mldsa-artifacts}}.
 
-A composite ML-DSA + EdDSA signature MUST use a hash algorithm with a digest size of at least 256 bits.
+A composite ML-DSA + EdDSA signature MUST use a hash algorithm with a digest size of at least 256 bits for the computation of the message digest.
 A verifying implementation MUST reject any composite ML-DSA + EdDSA signature that uses a hash algorithm with a smaller digest size.
 
 ### Key Material Packets
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -896,7 +896,6 @@ This has the advantage of an enhanced side-channel resistance of the signature o
 This specification requires that all PQ(/T) signatures defined in this document are made on message digests computed with a hash algorithm with at least 256 bits of digest size.
 Since all signature algorithms defined in this document require signature packets of version 6 or higher, which currently include a leading random salt value in the hashed data, hash-collision attacks do not apply to the signature schemes defined in this specification.
 Therefore, a hash algorithm with a digest size of 256 bits is sufficient to match the targeted security level of all PQ(/T) algorithms defined in this document, as it achieves a preimage and second preimage security of 256 bits.
-In such cases, the relevant security property is collision resistance.
 
 ## Symmetric Algorithms for SEIPD Packets
 
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -893,7 +893,7 @@ This has the advantage of an enhanced side-channel resistance of the signature o
 
 ## Minimum digest size for PQ(/T)-signatures
 
-This specification requires that all PQ(/T) signatures defined in this document are made in combination with a hash algorithm of at least 256 bits digest size.
+This specification requires that all PQ(/T) signatures defined in this document are made on message digests computed with a hash algorithm with at least 256 bits of digest size.
 The relevant security property is preimage resistance, since all signature algorithms defined in this document require signature packets of version 6 or higher, which currently include a random salt prefix.
 Therefore, a hash algorithm with 256 bits digest size is sufficient to match the targeted security level of all PQ(/T) algorithms defined in this document.
 Note that the minimum hash algorithm digest size requirement is different for signature schemes that are defined in [RFC9580], where signature algorithms are also allowed to be used with non-salted v4 signatures.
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -735,6 +735,15 @@ The algorithm-specific v6 signature parameters for ML-DSA + EdDSA signatures con
 
  - A fixed-length octet string of the ML-DSA signature value, whose length depends on the algorithm ID as specified in {{tab-mldsa-artifacts}}.
 
+A composite ML-DSA + EdDSA signature MUST use a hash algorithm whose digest size meets or exceeds the minimum digest size specified for the signature algorithm in {{tab-min-mldsa-eddsa-digest-size}}.
+A verifying implementation MUST reject any composite ML-DSA + EdDSA signature that uses a hash algorithm with a smaller digest size.
+
+{: title="Minimum hash algorithm digest size for composite ML-DSA + EdDSA signatures in bits" #tab-min-mldsa-eddsa-digest-size}
+Signature algorithm | Minimum digest size
+:-------------------| ------------------
+ML-DSA-65+Ed25519   | 384
```
falko-strenzke, line 744: Maybe the algorithm IDs should be given in a first column.

falko-strenzke (approved):
I agree with this proposal. If the other authors come to the conclusion that the digest size should be reduced to 256 for ML-DSA-65, I could also live with that. I don't see that a digest size greater than the security level is relevant to v6 signatures at all. But I agree that it makes sense to align to what RFC 9580 specifies for EdDSA. In any case to have a minimum requirement is good to hedge against digest substition attacks (which require breaking the 2nd-preimage resistance).

draft-ietf-openpgp-pqc.md
```diff
@@ -735,6 +735,15 @@ The algorithm-specific v6 signature parameters for ML-DSA + EdDSA signatures con
 
  - A fixed-length octet string of the ML-DSA signature value, whose length depends on the algorithm ID as specified in {{tab-mldsa-artifacts}}.
 
+A composite ML-DSA + EdDSA signature MUST use a hash algorithm whose digest size meets or exceeds the minimum digest size specified for the signature algorithm in {{tab-min-mldsa-eddsa-digest-size}}.
+A verifying implementation MUST reject any composite ML-DSA + EdDSA signature that uses a hash algorithm with a smaller digest size.
+
+{: title="Minimum hash algorithm digest size for composite ML-DSA + EdDSA signatures in bits" #tab-min-mldsa-eddsa-digest-size}
+ID | Algorithm         | Minimum digest size
+:--|:------------------|--------------------
+30 | ML-DSA-65+Ed25519 | 384
```
wussler, line 744: Given this is a minimum requirement, I would enforce 256 bit. This matches the MUST bitsize also with the hashing algorithms.

Note that SHA-256 corresponds to security level 2, and SHA-384 to level 4. We are dealing here with level 3, and I value more interoperability and availability of SHA(2/3)-256.

Also, SHA3-384 is not defined in RFC9580, and therefore if someone wants to use SHA3 they would be bound to use SHA3-512.

Pre-image attacks on hashes seem to be quite unrealistic, given we're using only salted signatures.

falko-strenzke: I'm fine with going for SHA-256. However, regarding the justification: The salt doesn't help against pre-image attacks. But SHA-256 offers a 256 bit pre-image resistance, so there is still a security margin from the 192 bit level of the algorithm. On the list our justification should be that 
  a) the salt prevents hash-collision attacks anyway and that 
  b) the (2nd) pre-image resistance of SHA-256 is still higher than what is required for this level and 
  c) SHA-256 is MUST in RFC 9580

TJ-91: If we want to be consistent with RFC 9580 I would say going with 384 bits is the most natural choice that is easy to justify. As Falko pointed out: ML-DSA is the relevant longterm-secure component and then following RFC 9580 the requirement should be 384 bits. Most of the arguments for going with 256 bits could be applied to reducing the requirements for other signature algorithms as well, so it doesn't seem very logical to me to use them only for this case.

That there is no SHA3-384 is unfortunate but IMO it's not a big deal. I would opt for 384 bits, or, we define our own but consistent reasoning for all 5 signature schemes (like Stavros suggests with NIST CSOR)

twiss: IMHO, using SHA256 could be justified by saying:

- For ML-DSA-65+Ed25519, we target security level 2, _but_ we are less confident in ML-DSA than the other components, due to it being a newer algorithm, therefore we bump that component to the next security level (without claiming that we actually achieve the higher security level, due to the other components not being bumped, and aforementioned uncertainty)
- For ML-DSA-87+Ed448, we target security level 5, but there is no higher level of ML-DSA so we didn't bump it
  (it's perhaps also debatable whether Ed448 achieves it, so maybe it's more accurate to say that this one sits between security level 4 and 5?)
- In SLH-DSA we have more confidence that it matches the stated security level, therefore we match it straightforwardly (as this PR already does)

For me, this would also be most consistent with RFC9580, which requires 256 bits for Ed25519 and 512 bits for Ed448.

TJ-91:
I changed the text to now require 256 bits instead. I welcome improvements to the text since I was struggling a bit to find a good wording.

Also I noticed we have two things we might have to address / circumvent somehow:

1. Technically, we can't speak about signature packets >v6 and claim they will have a random salt. How could we avoid this?
2. Using 256 bits for signature algorithms where 128/192 bits would suffice is not motivated. Since we do not have any such hash algorithms in OpenPGP (at least currently) it seems a bit silly to make the distinction, but of course new hash algorithms could be introduced at any time. What do you prefer here?

draft-ietf-openpgp-pqc.md
```diff
@@ -885,6 +891,14 @@ The algorithm ID identifies unequivocally the algorithm, the parameters for its
 This specification makes use of the default "hedged" variants of ML-DSA and SLH-DSA, which mix fresh randomness into the respective signature-generation algorithm's internal hashing step.
 This has the advantage of an enhanced side-channel resistance of the signature operations according to  {{FIPS-204}} and {{FIPS-205}}.
 
+## Minimum digest size for PQ(/T)-signatures
+
+This specification requires that all PQ(/T) signatures defined in this document are made in combination with a hash algorithm of at least 256 bits digest size.
+The relevant security property is preimage resistance, since all signature algorithms defined in this document require signature packets of version 6 or higher, which include a random salt value.
```
wussler, line 897:
```suggestion
The relevant security property is preimage resistance, since all signature algorithms defined in this document require signature packets of version 6 or higher, which currently include a random salt prefix.
```

wussler:
> Technically, we can't speak about signature packets >v6 and claim they will have a random salt. How could we avoid this?

I added a currently, that might be enough

draft-ietf-openpgp-pqc.md
```diff
@@ -885,6 +891,14 @@ The algorithm ID identifies unequivocally the algorithm, the parameters for its
...
+Therefore, a hash algorithm with 256 bits digest size is sufficient to match the targeted security level of all PQ(/T) algorithms defined in this document.
+Note that the minimum hash algorithm digest size requirement is different for signature schemes that are defined in [RFC9580], where signature algorithms are also allowed to be used with non-salted v4 signatures.
```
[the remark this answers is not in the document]
wussler: This sounds more like a "mailing list justification" rather than something that should live in the draft, but I am not strongly opposed to it

wussler (approved):

draft-ietf-openpgp-pqc.md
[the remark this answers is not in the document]
fluppe2: I would also be in favour of dropping this.

draft-ietf-openpgp-pqc.md
```diff
@@ -885,6 +891,14 @@ The algorithm ID identifies unequivocally the algorithm, the parameters for its
 This specification makes use of the default "hedged" variants of ML-DSA and SLH-DSA, which mix fresh randomness into the respective signature-generation algorithm's internal hashing step.
 This has the advantage of an enhanced side-channel resistance of the signature operations according to  {{FIPS-204}} and {{FIPS-205}}.
 
+## Minimum digest size for PQ(/T)-signatures
+
+This specification requires that all PQ(/T) signatures defined in this document are made on message digests computed with a hash algorithm with at least 256 bits of digest size.
+Since all signature algorithms defined in this document require signature packets of version 6 or higher, which currently include a leading random salt value in the hashed data, hash-collision attacks do not apply to the signature schemes defined in this specification.
+Therefore, a hash algorithm with a digest size of 256 bits is sufficient to match the targeted security level of all PQ(/T) algorithms defined in this document, as it achieves a preimage and second preimage security of 256 bits.
```
fluppe2, line 898:
```suggestion
Since all signature algorithms defined in this document require version 6 signature packets, which include a random salt in the hashed data, the required property is not collision but (2nd) preimage resistance. Therefore, a hash algorithm with a digest size of 256 bits is sufficient to match the targeted security level of all PQ(/T) algorithms defined in this document.
```
Simplified and dropped the 'v6 or higher' as we are specifying against existing versions. Also dropped mentioning 256 bits security since it might be confusing to people that this implicitely means classical attacks and I guess we would like to avoid citing NIST security category 5.

TJ-91: > include a random salt

should be 

> include a **leading** random salt **value**

---

> match the targeted security level of all PQ(/T) algorithms

I think should be

> match the targeted security level**s** of all PQ(/T) algorithms

---

Also, we have one line per sentence :)
Other than that I am fine with this (it's closer to my original text again without Falko's additions)

fluppe2: Ok, I added your suggestions.

PR #195, opened 2025-05-28
KEK combiner takes ECDH pubkey and ciphertext as input

dkg:
This just aligns the textual description with the pseudocode.


fluppe2 (approved):
@dkg Good catch!

PR #204, opened 2025-05-29
Normalize text in Key Generation Procedure sections

dkg:
It's not clear why the ML-* key generation sentences were sigificantly different from the ECC key generation sentences.

It was also not clear to me what "the relative specification" meant, so i've dropped that term.

The sections referenced don't seem to actually "specify formats", rather they describe the sizes of the objects, so i've adjusted that text too.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -699,7 +699,7 @@ Algorithm ID reference | ML-DSA    | Public key | Secret key | Signature value
 
 The implementation MUST generate the ML-DSA and the EdDSA component keys independently.
 ML-DSA key generation follows the specification in [FIPS-204], and the artifacts are encoded as fixed-length octet strings whose sizes are listed in {{mldsa-signature}}.
-EdDSA key generation follows the specification in {{RFC7748}}, and the artifacts are encoded as fixed-length octet strings whose sizes are listed in {{eddsa-signature}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -698,8 +698,8 @@ Algorithm ID reference | ML-DSA    | Public key | Secret key | Signature value
 ### Key generation procedure {#ecc-mldsa-generation}
 
 The implementation MUST generate the ML-DSA and the EdDSA component keys independently.
-ML-DSA key generation follows the specification [FIPS-204] and the artifacts are encoded as fixed-length octet strings as defined in {{mldsa-signature}}.
-For EdDSA this is done following the relative specification in {{RFC7748}}, and encoding the artifacts as specified in {{eddsa-signature}} as fixed-length octet strings.
+ML-DSA key generation follows the specification in [FIPS-204], and the artifacts are encoded as fixed-length octet strings whose sizes are listed in {{mldsa-signature}}.
+EdDSA key generation follows the specification in {{RFC7748}}, and the artifacts are encoded as fixed-length octet strings whose sizes are listed in {{eddsa-signature}}.
```
fluppe2, line 702:
```suggestion
EdDSA key generation follows the specification in {{RFC8032}}, and the artifacts are encoded as fixed-length octet strings whose sizes are listed in {{eddsa-signature}}.
```

dkg: good catch, @fluppe2 .  I've fixed it now.

fluppe2 (approved):

TJ-91 (approved):
LGTM

wussler (approved):

PR #205, opened 2025-05-29
Correct octet lengths in the IANA table

dkg:
This fixes three things:

- uses the seed-style secret key for ML-KEM
- uses the seed-style secret key for ML-DSA
- updates the lengths of the ML-DSA signatures

I think i've got it right here, but please double-check!


dkg:
I've also added a patch that corrects the order of the octets in the final PKESK.

wussler (approved):
Double checked all the sizes, looks correct to me

PR #207, opened 2025-06-02
remove superfluous field size from tables (issue #193)

### LINKED ISSUE #193: why do the ECDH and EdDSA tables need the "Field size" property?
I don't see any reference to "field size" anywhere else in the document, but it is present in the tables labeled `tab-ecdh-cfrg-artifacts` and `tab-eddsa-artifacts`.  can we drop it?

Field size is mentioned in [§9.2 of RFC 9580](https://www.rfc-editor.org/rfc/rfc9580.html#section-9.2-4) but only because that specification is juggling the historic complexity introduced by the different possible formats for point formats in the OID-parameterized "ECDH" and "ECDSA" legacy from RFC 6637.

I don't think "field size" is relevant for the PQC document.

TJ-91:
resolves #193


fluppe2 (approved):
LGTM

PR #208, opened 2025-06-02
Trim mlkem ops

### LINKED ISSUE #194: Trim `mlkem-ops` section
The subsection labeled `mlkem-ops` contains what appears to be superfluous text:

```
The procedure to perform `ML-KEM.Encaps()` is as follows:

 1. Invoke `(mlkemCipherText, mlkemKeyShare) <- ML-KEM.Encaps(mlkemPublicKey)`, where `mlkemPublicKey` is the recipient's public key

 2. Set `mlkemCipherText` as the ML-KEM ciphertext

 3. Set `mlkemKeyShare` as the ML-KEM symmetric key share

The procedure to perform `ML-KEM.Decaps()` is as follows:

 1. Invoke `mlkemKeyShare <-  ML-KEM.Decaps(mlkemCipherText, mlkemSecretKey)`

 2. Set `mlkemKeyShare` as the ML-KEM symmetric key share
```

If it's not superfluous, then it seems tautological: To do `ML-KEM.Decaps()`, invoke `ML-KEM.Decaps()`. 😂

Can we just cut it?

TJ-91:
resolves #194

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -479,20 +479,6 @@ Algorithm ID reference | ML-KEM      | Public key | Secret key | Ciphertext | Ke
 
 To instantiate `ML-KEM`, one must select a parameter set from the column "ML-KEM" of {{tab-mlkem-artifacts}}.
 
-The procedure to perform `ML-KEM.Encaps()` is as follows:
-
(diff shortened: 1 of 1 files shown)


wussler (approved):

fluppe2 (approved):
LGTM

(IMHO once removed the text becomes more consistent, in particular when looking at the straightforward EdDSA and ML-DSA description)

Issue #199, opened 2025-05-29
is it "EC" or "ECC"?


dkg:
The current document, when talking about material related to elliptic curve cryptography says:

- EC point public key
- EC public key

But also:

- ECC Requirements
- ECC data structure
- ECC key
- ECC signature
- ECC key derivation function

Is the reader supposed to infer something different from "EC" versus "ECC"?  if not, can we pick either EC or ECC and use it consistently throughout?

TJ-91:
I think we can just use "ECC" everywhere. However, to me it seems quite normal to use both terms. RFC 9580 also uses both.

I will create a PR that changes EC to ECC and we can decide if we want to go with that.

fluppe2:
Closed with #210

PR #210, opened 2025-06-02
use ECC instead of EC consistently

TJ-91:
addresses #199

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -631,7 +631,7 @@ The composite ML-KEM-1024 + X448 (algorithm ID 36) MUST be used only with v6 key
 
 The algorithm-specific public key is this series of values:
 
- - A fixed-length octet string representing an EC point public key, in the point format associated with the curve specified in {{ecc-kem}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -631,7 +631,7 @@ The composite ML-KEM-1024 + X448 (algorithm ID 36) MUST be used only with v6 key
 
 The algorithm-specific public key is this series of values:
 
- - A fixed-length octet string representing an EC point public key, in the point format associated with the curve specified in {{ecc-kem}}.
+ - A fixed-length octet string representing an ECC point public key, in the point format associated with the curve specified in {{ecc-kem}}.
```
fluppe2, line 634:
```suggestion
 - A fixed-length octet string representing an ECC public key, in the point format associated with the curve specified in {{ecc-kem}}.
```
There is no such thing as an "ECC point".

fluppe2 (changes requested):
Apart from the one "ECC point public key" -> "ECC public key", LGTM.

wussler (approved):

fluppe2 (approved):
GTG

Issue #202, opened 2025-05-29
names of SLH functions


dkg:
The document currently claims that the functions are named `SLH-DSA.sign` and `SLH-DSA.verify`, but those names are not present in FIPS 205.

While [FIPS204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf) does name its functions `ML-DSA.sign` and `ML-DSA.verify`, [FIPS205](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.205.pdf) appears to name its functions `slh_sign` and `slh_verify` (or `hash_slh_sign` and `_hash_slh_verify`, but i don't think OpenPGP is aiming to use those).

Where this document refers to functions found in FIPS 205, it should use the literal function names that appear in that documet.

TJ-91:
resolved by #211

Issue #200, opened 2025-05-29
normalize order of arguments `EdDSA.Verify` and `ML-DSA.Verify`


dkg:
The current draft says:

-  `EdDSA.Verify(eddsaPublicKey, eddsaSignature, dataDigest)`
- `ML-DSA.Verify(mldsaPublicKey, dataDigest, mldsaSignature)`

Why are the final two parameters in different positions?

In [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032) there is no formal declaration of `EdDSA.Verify`, but [the python example](https://datatracker.ietf.org/doc/html/rfc8032#page-23) says `verify(public, message, signature)`, which is not the order that appears here.

falko-strenzke:
Should be resolved by https://github.com/openpgp-pqc/draft-openpgp-pqc/pull/211/files

TJ-91:
indeed, I did not properly format the text for auto closing on merge

PR #211, opened 2025-06-02
Clarity context parameter; rename SLH-DSA algorithms to match FIPS 205; change input order of EdDSA.Verify

### LINKED ISSUE #201: signature context parameters
The document says `ML-DSA.Verify(mldsaPublicKey, dataDigest, mldsaSignature)`, but [FIPS204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf) declares `ML-DSA.Verify(𝑝𝑘, 𝑀, 𝜎, 𝑐𝑡𝑥)`, where `𝑀` is the message, `𝜎` is the signature, and `𝑐𝑡𝑥` is the context.

Likewise, the document says `ML-DSA.Sign(mldsaSecretKey, dataDigest)`, but FIPS204 declares `ML-DSA.Sign(𝑠𝑘, 𝑀, 𝑐𝑡𝑥)`.

Both Ed448 and Ed25519 as defined in [RFC8032](https://datatracker.ietf.org/doc/html/rfc8032) have context parameters, as does `slh_verify` and `slh_sign` in [FIPS205](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.205.pdf).

But the term `context` comes up only twice in the document, and not about signatures at all.  And the string `ctx` appears nowhere either.

I think this means that the context is always the empty string.  It might be nice to say that explicitly somewhere.

TJ-91:
resolves #201, #202, #200

I chose to do this in one PR since it's related and to avoid conflicts.

I am not sure about repeating the statement about the context, but I felt it's most clear to mention it when introducing each of the algorithms.


draft-ietf-openpgp-pqc.md
```diff
@@ -679,12 +680,12 @@ Algorithm ID reference | Curve   | Field size | Public key | Secret key | Signat
 
 Throughout this specification ML-DSA refers to the default pure and hedged version of ML-DSA defined in [FIPS-204].
 
-For ML-DSA signature generation the default hedged version of the algorithm `ML-DSA.Sign` given in [FIPS-204] is used.
+For ML-DSA signature generation, the default hedged version of the `ML-DSA.Sign` algorithm, as specified in [FIPS-204], is used with an empty context string `ctx`.
```
wussler, line 683:
```suggestion
For ML-DSA signature generation the default hedged version of the `ML-DSA.Sign` algorithm is used with an empty context string `ctx`, as specified in [FIPS-204].
```

I feel like with less subordinate clauses it reads a bit smoother

PR #212, opened 2025-06-03
identify signature type for subkey binding signature

### LINKED ISSUE #206: Identify signature types
In "**Hash Algorithms for Key Binding Signatures**", it says "subkey binding signatures" and "primary key binding signatures".  Perhaps it would be cleaner to include a parenthetical aside like "(signature type 0x18)" in the relevant place.

Note that #198 probably needs to be resolved first.

TJ-91:
resolves #206


fluppe2 (approved):
LGTM

PR #216, opened 2025-06-09
FIPS references should be normative

dkg:
No implementer can implement this draft without either the mathematics or the serialization formats described in the FIPS documents.  They are effectively normative.


fluppe2 (approved):
Fine for me

PR #218, opened 2025-06-09
Include the IANA specification commentary as notes for the field.

dkg:
Looking at the IANA table at
https://www.iana.org/assignments/openpgp/openpgp.xhtml#openpgp-public-key-algorithms i notice that X25519 and X448 PKESK entries should also be updated to include the comparable fields, but that doesn't need to happen here.


dkg:
I offer this just so it's clear precisely what we're asking IANA to do.  the textual explanation doesn't mean anything to IANA -- but it should mean something to someone looking at the registry.  so explicitly tell IANA to put it in the notes field for the registry.

The fact that X25519 and X448's PKESK column isn't spelled out as clearly in the registry is a separate bug in the registry, which is not the job of this draft to fix.

wussler (approved):

fluppe2 (approved):

TJ-91 (approved):

PR #223, opened 2025-07-01
fixed typo in IANA table (missing word 'key')

falko-strenzke:


fluppe2 (approved):
I fixed some further typos and switched order of inputs in ECDH-KEM.Decaps() to match the order of ML-KEM.Decaps().

wussler (approved):

PR #227, opened 2025-09-15
Migration: explicitly discuss verification separately from signing

dkg:


draft-ietf-openpgp-pqc.md
```diff
@@ -860,6 +860,11 @@ An implementation should not abort the encryption process when encrypting a mess
 An implementation may sign with both a PQ(/T) and a traditional key using multiple signatures over the same data as described in {{multiple-signatures}}.
 Signing only with PQ(/T) key material is not backwards compatible.
 
+## Verifying with Traditional and PQ(/T) Keys
+
+When verifying, an implementation MAY be willing to accept signatures PQ(/T) keys and from traditional keys.
+A verifier particularly concerned with a cryptographically relevant quantum computer with knowledge of a peer that has a PQ(/T) signing key MAY prefer instead to ignore all traditional signatures from that peer.
```
fluppe2, line 866:
```suggestion
A verifier particularly with knowledge of a peer that has a PQ(/T) signing key MAY prefer instead to ignore all traditional signatures from that peer.
```
I think it is ok if one deletes the mentioning of being concerned with a CRQC.

dkg: I agree that there is no formal need for this statement.  Verifiers can already make whatever policy decisions they want to (like declining to verify signatures from DSA 1024).

I'm proposing this because i think it encourages implementers to think about migration from all three of these directions, not just the two (signing and encrypting).

I personally prefer to keep some sort of plausible motivational guidance ("concerned about a cryptographically relevant quantum computer") rather than just saying "hey, verifiers might do this thing".  I'm open to other motivational justifications, of course.

wussler: I see your point @dkg but would still drop the `particularly`

PR #242, opened 2025-09-18
addressed nits from AD Review (#237)

### LINKED ISSUE #237: AD Review: Comments / NITS
AD review: https://mailarchive.ietf.org/arch/msg/openpgp/zhAL7mFI-p3-7njGsEVkqyIZsCc/

Comments / NITS
 
Abstract:

>  	advent of quantum computers
> 
>  	in the presence of quantum computers.
 
Can we use the same terminology as PQUIP recommends? eg here we should use
cryptographically relevant quantum computer (CRQC) and "protect against CRQCs"
 
(see https://datatracker.ietf.org/doc/html/draft-ietf-pquip-pqc-engineers)

 
 Introduction:
 
>  	Such cryptographic algorithms are referred to as post-quantum cryptography. Th
 
Perhaps introduce the acronym here, too eg post-quantum cryptography (PQC) to align it
with PQUIP?

 
>  	For the two ML-* schemes, this document follows the conservative
>  	strategy to deploy post-quantum in combination with traditional
>  	schemes such that the security is retained even if all schemes
>  	but one in the combination are broken.
 
Maybe add a sentence that this is called a hybrid? (you do refer to I-D.ietf-pquip-pqt-hybrid-terminology
later on int the document but not yet here)
 
 
>  	from the respective elliptic curve specification.
 
Can this get a normative reference appended?

falko-strenzke:
Closes #237

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -229,7 +229,8 @@ Such cryptographic algorithms are referred to as post-quantum cryptography (PQC)
 The algorithms defined in this extension were chosen for standardization by the National Institute of Standards and Technology (NIST) in mid 2022 {{NISTIR-8413}} as the result of the NIST Post-Quantum Cryptography Standardization process initiated in 2016 {{NIST-PQC}}.
 Namely, these are ML-KEM {{FIPS-203}} as a Key Encapsulation Mechanism (KEM), a KEM being a modern building block for public-key encryption, and ML-DSA {{FIPS-204}} as well as SLH-DSA {{FIPS-205}} as signature schemes.
 
-For the two ML-* schemes, this document follows the conservative strategy to deploy post-quantum in combination with traditional schemes such that the security is retained even if all schemes but one in the combination are broken. Such combinations are referred to as "multi-algorithm" "post-quantum/traditional" (PQ/T) hybrid algorithms.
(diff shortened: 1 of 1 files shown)


draft-ietf-openpgp-pqc.md
```diff
@@ -223,13 +223,14 @@ Specifically, it defines composite public-key encryption based on ML-KEM (former
...
 
-Such cryptographic algorithms are referred to as post-quantum cryptography.
+Such cryptographic algorithms are referred to as post-quantum cryptography (PQC).
 The algorithms defined in this extension were chosen for standardization by the National Institute of Standards and Technology (NIST) in mid 2022 {{NISTIR-8413}} as the result of the NIST Post-Quantum Cryptography Standardization process initiated in 2016 {{NIST-PQC}}.
 Namely, these are ML-KEM {{FIPS-203}} as a Key Encapsulation Mechanism (KEM), a KEM being a modern building block for public-key encryption, and ML-DSA {{FIPS-204}} as well as SLH-DSA {{FIPS-205}} as signature schemes.
 
 For the two ML-* schemes, this document follows the conservative strategy to deploy post-quantum in combination with traditional schemes such that the security is retained even if all schemes but one in the combination are broken.
+Such combinations are referred to as multi-algorithm "post-quantum/traditional" (PQ/T) hybrid algorithms.
```
falko-strenzke, line 233:
```suggestion
Such combinations are referred to as multi-algorithm or "post-quantum/traditional" (PQ/T) hybrid algorithms.
```

fluppe2 (approved):
LGTM

PR #253, opened 2025-12-10
Address Éric's review

wussler:
https://mailarchive.ietf.org/arch/msg/openpgp/TTdgDRqqHFMY4tt2FUMI46pzMIw/

PR #255, opened 2025-12-13
Handle update tag/info

fluppe2:
I chose to remove the update tag for reasons of simplicity and since #253 adds a reference to RFC9580 to the abstract, so the setting and nature of this pqc spec should be clear.

Alternatively, I am open to add the tag back in again and add a clarifying sentence to the abstract.


TJ-91:
I am fine with removing the tag, but I suggest to rephrase #253 in that regard. We only added `updating [RFC9580]` to contextualize the update tag, right? It seems a bit odd to me to not have the update tag, but in the abstract write about updating RFC9580.

My proposal: We can either remove it, or replace "updating" by "extending" or something similar that marks it as an extension to RFC9580.

PR #256, opened 2025-12-15
Address Deb's review

wussler:
https://mailarchive.ietf.org/arch/msg/openpgp/Uj0HMET-N6Pyz3BBoddGUss9ds8/


fluppe2 (commented):
LGTM

What about the aspects:

- Section 3.3, last para: (why not MUST) Under what circumstances would an
implementation not consider a message correctly signed?

- Section 4.2, second to last bullet: The session key is generated by the sender?

fluppe2 (approved):

PR #262, opened 2025-12-24
Convert MUST/SHOULD/MAY from algorithm tables to text

dkg:
Éric Vyncke's DISCUSS specifically asked for these MTI indicators to be moved out of the table and into the text.

It's not clear to me that this changes the meaning of the draft in any clear way, which I suppose is both an argument for and against making this change.

### DIFF
--- a/draft-ietf-openpgp-pqc.md
+++ b/draft-ietf-openpgp-pqc.md
@@ -354,7 +354,6 @@ The SLH-DSA algorithms offer an additional performance tradeoff between signatur
 The larger parameter sets of ML-DSA and ML-KEM (Algorithm IDs 31 and 36) are recommended to support interoperability, but they are not required for compliance.
 Implementations targeting highly constrained environments may omit these larger variants.
 
-SLH-DSA is also offered in a "fast" and a "small" variant to allow for further tradeoffs.
 For SLH-DSA-SHAKE-256, only the "small" variant is offered to contain signature size.
 See also {{performance-considerations}} for further considerations about parameter choices.
 


draft-ietf-openpgp-pqc.md
```diff
@@ -327,24 +327,30 @@ All of these schemes are fully specified via their algorithm ID, that is, they a
...
+36 | ML-KEM-1024+X448                 | {{ecc-mlkem}}
 
 
+A conformant implementation MUST implement ML-DSA-65+Ed25519 and ML-KEM-768+X25519.
+It SHOULD also implement the ML-DSA-87+Ed448 and ML-KEM-1024+X448,
+but may omit them if targeting a highly constrained environment.
+An implementation MAY implement any of the SLH-DSA algorithms.
+
 The specified algorithm IDs offer two security levels for each scheme, for a tradeoff between security and performance.
+he SLH-DSA algorithms offer an additional performance tradeoff between signature generation time ("128f" is faster) and signature size ("128s" is smaller) at the lower of the two SLH-DSA security levels.
```
fluppe2:
```suggestion
The SLH-DSA algorithms offer an additional performance tradeoff between signature generation time ("128f" is faster) and signature size ("128s" is smaller) at the lower of the two SLH-DSA security levels.
```
Typo. Otherwise LGTM.

dkg: d'oh!  thank you ☺  fixed now.

fluppe2:
@TJ-91 After the removed (vague) sentence we still point the reader to Section 10.1 that discusses performance aspects in more detail. I'm fine with @dkg's proposal as it is right now.

@TJ-91 If you are fine also, please merge.

